PR #35, opened 2018-02-14
Single iteration method

SteveMacenski:


SteveMacenski:
Major thread of move base down from ~80-120 nominal for 6 sensors down to 20-40%

PR #74, opened 2018-07-19
clearing buffer after taking current measurements from the buffer to …

SteveMacenski:
…not over publish the same infromation. Addresses #72


SteveMacenski:
I don't quite understand what you mean. I dont include buffer in the parameter name because its too many words. All the layer references are to that parameterized name. 
Later in the buffer its read into saying buffer to give the reader of that code a deeper understanding of what that means when tracing through the stack

tasilb:
In the example configs, it's not obvious what the parameter does, given the current name. That's ultimately my concern.


________________________________

SteveMacenski:
got it. I'll add more docs. I know internally we like having excessively long param names but as long as I add enough relevant docs it should be clear in context for the larger community involved.

PR #76, opened 2018-08-10
Refactor decay acceleration for mathematical legibility

tasilb:
Haven't completed testing, but feedback is appreciated

### DIFF
--- a/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
+++ b/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
@@ -137,8 +137,8 @@ protected:
   bool IsGridEmpty(void) const;
 
   // Get time information for clearing
-  double GetDecayShift(const double& time_delta);
-  double GetAcceleratedDecayShift(const double& time_delta, \
+  double GetBaseDurationToDecay(const double& time_delta);
+  double GetAcceleratedDecayModifier(const double& time_delta, \
                                   const double& acceleration_factor);
   void TemporalClearAndGenerateCostmap(std::vector<frustum_model>& frustums);
 
--- a/src/spatio_temporal_voxel_grid.cpp
+++ b/src/spatio_temporal_voxel_grid.cpp
@@ -150,7 +150,8 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
     bool frustum_cycle = false;
 
     const double time_since_marking = cur_time - cit_grid.getValue();
-    const double decay_shift = GetDecayShift(time_since_marking);
+    const double base_duration_to_decay = GetBaseDurationToDecay( \
+                                                            time_since_marking);
 
     for(frustum_it; frustum_it != frustums.end(); ++frustum_it)
     {
@@ -158,13 +159,12 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
       {
         frustum_cycle = true;
 
-        // accel_decay_shift is the amount of time from the current time until
-        // the time a given mark is supposed to be cleared
-        const double accel_decay_shift = \
-          GetAcceleratedDecayShift(time_since_marking, \
-                                   frustum_it->accel_factor);
+        const double decay_acceleration = GetAcceleratedDecayModifier( \
+                                  time_since_marking, frustum_it->accel_factor);
 
-        if (accel_decay_shift < 0)
+        const double time_until_decay = base_duration_to_decay - \
+          decay_acceleration;
+        if (time_until_decay <= 0)
         {
           // expired by acceleration
           if(!this->ClearGridPoint(pt_index))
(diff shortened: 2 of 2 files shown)


include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
```diff
@@ -137,8 +137,8 @@ class SpatioTemporalVoxelGrid
   bool IsGridEmpty(void) const;
 
   // Get time information for clearing
-  double GetDecayShift(const double& time_delta);
-  double GetAcceleratedDecayShift(const double& time_delta, \
+  double GetBaseDurationToDecay(const double& time_delta);
```
SteveMacenski: This is just the standard temporal expiration. Perhaps call it GetTemporalClearingTime?

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
```diff
@@ -137,8 +137,8 @@ class SpatioTemporalVoxelGrid
...
+  double GetAcceleratedDecayModifier(const double& time_delta, \
```
SteveMacenski: This is the acceleration factor, GetAccelerationDuration? or something along those lines

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -174,8 +174,7 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
         }
         else
         {
-          const double updated_mark = cit_grid.getValue() - \
-            (decay_shift - accel_decay_shift);
+          const double updated_mark = cit_grid.getValue() - decay_acceleration;
```
SteveMacenski: decay_acceleration -> accelerated_duration or something similar. acceleration and decay are both in the same vein

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
tasilb: `acceleration_factor` is already used as a parameter for this function--I'm not opposed to changing both if you have suggestions

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
tasilb: How about `GetTemporalClearingDuration`? My intention is to make it clear to readers that what it returns is the length of time, not a single timestamp (which was what was done before acceleration was implemented)

src/spatio_temporal_voxel_grid.cpp
tasilb: How about just `acceleration`?

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
SteveMacenski: ok

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
SteveMacenski: Acceleration duration?

src/spatio_temporal_voxel_grid.cpp
SteveMacenski: see comment above

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
tasilb: How about rename `GetAcceleratedDecayModifier -> GetAccelerationComponent` because it's not specifically intended to be used as a duration--it describes an influence on the resulting decay duration.

src/spatio_temporal_voxel_grid.cpp
tasilb: see comment above

tasilb:
Passes testing

src/spatio_temporal_voxel_grid.cpp
SteveMacenski: see comment above

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
SteveMacenski: The acceleration itself is the component, its not a modifier or anything, its what I would consider to be the acceleration term. GetFrustumAcceleration ?

tasilb: I agree, and that was the intent--I wanted to use the word "term" but that seemed a little too mathy. I like `GetFrustumAcceleration`

src/spatio_temporal_voxel_grid.cpp
tasilb: see comment above

SteveMacenski: see comment above

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
SteveMacenski: :+1: do itt

PR #81, opened 2018-10-08
first melodic migration attempt

doisyg:
Just made the appropriate changes for the package to compile and the minimal_test.launch to run without error. No test on real platform nor simulation yet done yet

### DIFF
--- a/include/spatio_temporal_voxel_layer/measurement_buffer.hpp
+++ b/include/spatio_temporal_voxel_layer/measurement_buffer.hpp
@@ -52,21 +52,16 @@
 #include <ros/ros.h>
 #include <ros/time.h>
 // TF
-#include <tf/transform_listener.h>
-#include <tf/transform_datatypes.h>
-// msgs
-#include <sensor_msgs/PointCloud2.h>
-#include <geometry_msgs/Quaternion.h>
-// Mutex
-#include <boost/thread.hpp>
-
-#include <geometry_msgs/PoseStamped.h>
-
 #include "tf2_ros/transform_listener.h"
 #include "tf2_ros/message_filter.h"
 #include "message_filters/subscriber.h"
 #include "tf2_geometry_msgs/tf2_geometry_msgs.h"
-#include <tf2/buffer_core.h>
+// msgs
+#include <sensor_msgs/PointCloud2.h>
+#include <geometry_msgs/Quaternion.h>
+#include <geometry_msgs/PoseStamped.h>
+// Mutex
+#include <boost/thread.hpp>
 
 namespace buffer
 {
--- a/include/spatio_temporal_voxel_layer/measurement_buffer.hpp
+++ b/include/spatio_temporal_voxel_layer/measurement_buffer.hpp
@@ -1,4 +1,4 @@
-/*********************************************************************
+/********************************************************************
  *
  * Software License Agreement
  *
@@ -60,6 +60,14 @@
 // Mutex
 #include <boost/thread.hpp>
 
+#include <geometry_msgs/PoseStamped.h>
+
+#include "tf2_ros/transform_listener.h"
+#include "tf2_ros/message_filter.h"
+#include "message_filters/subscriber.h"
+#include "tf2_geometry_msgs/tf2_geometry_msgs.h"
+#include <tf2/buffer_core.h>
+
 namespace buffer
 {
 
@@ -77,7 +85,7 @@ public:
                     const double& min_obstacle_height,   \
                     const double& max_obstacle_height,   \
                     const double& obstacle_range,        \
-                    tf::TransformListener& tf,           \
+                    tf2_ros::Buffer& tf,                 \
                     const std::string& global_frame,     \
                     const std::string& sensor_frame,     \
                     const double& tf_tolerance,          \
@@ -113,7 +121,7 @@ private:
   // Removing old observations from buffer
   void RemoveStaleObservations(void);
 
-  tf::TransformListener& _tf;
+  tf2_ros::Buffer& _buffer;
   const ros::Duration _observation_keep_time, _expected_update_rate;
   boost::recursive_mutex _lock;
   ros::Time _last_updated;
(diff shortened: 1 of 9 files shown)


src/spatio_temporal_voxel_layer.cpp
```diff
@@ -188,7 +188,7 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
         boost::shared_ptr <buffer::MeasurementBuffer>
         (new buffer::MeasurementBuffer(topic, observation_keep_time,     \
         expected_update_rate, min_obstacle_height, max_obstacle_height,  \
-        obstacle_range, *tf_, _global_frame,                             \
+        obstacle_range, *tf_,  _global_frame,                            \
```
SteveMacenski: Looks like you may have added an extra space here.

src/measurement_buffer.cpp
```diff
@@ -149,7 +148,8 @@ void MeasurementBuffer::BufferPCLCloud(const \
 
     point_cloud_ptr cld_global(new pcl::PointCloud<pcl::PointXYZ>);
 
-    pcl_ros::transformPointCloud(_global_frame, cloud, *cld_global, _tf);
+    //pcl_ros::transformPointCloud(_global_frame, cloud, *cld_global, _tf);
```
SteveMacenski: You can just remove this line entirely

include/spatio_temporal_voxel_layer/measurement_buffer.hpp
```diff
@@ -60,6 +60,14 @@
 // Mutex
 #include <boost/thread.hpp>
 
+#include <geometry_msgs/PoseStamped.h>
+
+#include "tf2_ros/transform_listener.h"
```
SteveMacenski: Please add these headers under "TF" (and you can rename to TF2 if you like)

include/spatio_temporal_voxel_layer/measurement_buffer.hpp
```diff
@@ -60,6 +60,14 @@
 // Mutex
 #include <boost/thread.hpp>
 
+#include <geometry_msgs/PoseStamped.h>
```
SteveMacenski: Please add these headers under "msgs"

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
```diff
@@ -70,6 +70,12 @@
 #include <tf/message_filter.h>
 #include <tf/exceptions.h>
 
+#include "tf2_ros/transform_listener.h"
```
SteveMacenski: Also put these headers in the appropriate bin please :)

src/measurement_buffer.cpp
```diff
@@ -71,7 +71,8 @@ MeasurementBuffer::MeasurementBuffer(const std::string& topic_name,       \
     _vertical_fov(vFOV), _horizontal_fov(hFOV),
     _decay_acceleration(decay_acceleration), _marking(marking),
     _clearing(clearing), _voxel_size(voxel_size), _voxel_filter(voxel_filter),
-    _clear_buffer_after_reading(clear_buffer_after_reading)
+    _clear_buffer_after_reading(clear_buffer_after_reading),
+    _buffer(tf)
```
SteveMacenski: It looks like you just moved this down, can you put back up on the line 65 like it was before?

src/measurement_buffer.cpp
```diff
@@ -115,23 +116,21 @@ void MeasurementBuffer::BufferPCLCloud(const \
...
-    _tf.transformPose(_global_frame, local_pose, global_pose);
-
-    _observation_list.front()._origin.x = global_pose.getOrigin().getX();
-    _observation_list.front()._origin.y = global_pose.getOrigin().getY();
-    _observation_list.front()._origin.z = global_pose.getOrigin().getZ();
-
-    tf::quaternionTFToMsg(global_pose.getRotation(), orientation);
-    _observation_list.front()._orientation = orientation;
+    geometry_msgs::PoseStamped  local_pose, global_pose;
+    //local_pose.setOrigin(tf::Vector3(0, 0, 0));
```
SteveMacenski: Can you readd these lines? I'm a strong believer that a strong offense for stupid programming (for me) is a defense for myself.

SteveMacenski:
Ah, last thing, can you remove the TF1 dependencies and headers?

src/measurement_buffer.cpp
doisyg: Not sure what you meant but I removed the commented lines

doisyg:
I made the changes and replaced tf by tf2 in the package.xml and CMakeList + removed some headers. However, parts of the code (minimal_test.cpp for instance with #include <tf/transform_broadcaster.h>)
) are still using tf1 grammar. I am not sure I understand exactly what is going on, but I think tf2 is supporting the old tf1 grammar and did not deprecate it yet. Anyway this is a work in progress. I believe we should check if it can run properly on Melodic first, and then try to get rid completely of tf1 grammar.

SteveMacenski:
Sounds good, I'd appreciate if you could add back in my setRotation and setOrigin calls in the transformations. They're there to protect myself and others from doing accidental damage.

After that, I'll merge and open a new ticket to find a way to test this and then release to bloom. Have you run this code on your system and given the checkmark that it operates? Given the minor changes, I would assume if it worked at all for you its probably OK.

SteveMacenski:
Awesome, thanks for the hard work. 

Did you test this on your hardware perchance?

doisyg:
I initialized the value of the geometry_msgs::PoseStamped local_pose. Though it is verbose since I don't know any workaround for this old issue: https://github.com/ros/ros_comm/issues/148
The only check I did was to start minimal_test.launch. I never used this package before and was intended to try it on our platform (which we upraded to Melodic recently), hence the effort of building it. I am planning to try on our real robot asap but I am a bit short on time these days.

doisyg:
> 
> Did you test this on your hardware perchance?
I will report here once tested

SteveMacenski:
OK, feel free to reach back out if you find any issues. Given that its just a tf->tf2 conversion I would expect that if it worked at all its probably fine. 

I thought the API for class loader changed in melodic too so I'm a little surprised no pluginlib things needed updating here

SteveMacenski:
It is a `devel` branch after all ;-)

PR #83, opened 2018-10-09
allow for tracking unknown space to effect the default value in the grid

SteveMacenski:


SteveMacenski:
cherry picking to melodic

PR #86, opened 2018-10-17
[WIP] Dynamic reconfigure proposal

veimox:
Hi!

I have made a prototype of dynamic reconfigure for ease development. I do this PR as a discussion platform/proposal so we can add the changes later and, if you like it, merge.

The **motivation** is to have dynamic reconfigure to ease development and test faster.

The **current proposal** only offers the voxel\_grid and non-observation-sources-related parameters (but for the mapping related stuff). The reason is mostly that I wanted to propose something fast and that there is not that much need in changing the non-observation-sources-related params on runtime (but for the acceleration one). I also tried to follow tacit rules (e.g. no smart pointers) and initialize params from ros param. There was need of some ugly things (e.g. pointer of node handler, privatizing \_voxel\_decay and others) for not interfering so much with the current code. I think this has to change if we decide to merge.

I have tried to follow the code convention but I think it would be very nice if you add a .clang-format like file so all contributions are format-agnostic. I usually use [this one](https://github.com/ament/ament_lint/blob/master/ament_clang_format/ament_clang_format/configuration/.clang-format).

Things to discuss:
1. Does it make sense to have dynamic reconfigure?
2. Do observation sources need it too?

Last note: Thanks for the package!


SteveMacenski:
Hi,

thanks for the hard work. In general, I dislike dynamic reconfigure, mostly why I didn't include it. But hey, you did the work so there's no reason we can't add it!

I'm going to need to have some discussions over the remainder of this week to determine how to best approach this. In the meantime, I'll review some of the smaller formatting things I'd want changed and we can circle back to larger changes on the results of that discussion.

CMakeLists.txt
```diff
@@ -18,15 +18,20 @@ find_package(catkin REQUIRED COMPONENTS
...
   visualization_msgs
   message_generation
+  dynamic_reconfigure
 )
 
 add_service_files(DIRECTORY msgs
   FILES
   SaveGrid.srv
 )
+generate_dynamic_reconfigure_options(
```
SteveMacenski: Please add spaces before and after this macro.

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
```diff
@@ -43,9 +43,11 @@
 
 // voxel grid
 #include <spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp>
+#include <spatio_temporal_voxel_layer/SpatioTemporalVoxelLayerConfig.h>
```
SteveMacenski: Thanks for putting all the headers in their bins!

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
```diff
@@ -126,6 +130,11 @@ class SpatioTemporalVoxelLayer : public costmap_2d::CostmapLayer
   bool AddStaticObservations(const observation::MeasurementReading& obs);
   bool RemoveStaticObservations(void);
 
+  // Dynamic reconfigure
+  void DynamicReconfigureCallback(dynamicReconfigureType &config, uint32_t level);
+  dynamicReconfigureServerType* _dynamic_reconfigure_server;
+  bool _first_time;
```
SteveMacenski: I don't understand the need for this variable

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -62,15 +64,15 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
 
   // initialize parameters, grid, and sub/pubs
   ros::NodeHandle nh("~/" + name_), g_nh, prefix_nh;
+  _nh = &nh;
```
SteveMacenski: I don't understand the need for this (also its definition in the header needs to be tabbed over with the other variables. Why a pointer?)

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -112,18 +114,17 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
     default_value_ = costmap_2d::FREE_SPACE;
   }
 
-  if (_publish_voxels)
-  {
-    _voxel_pub = nh.advertise<sensor_msgs::PointCloud2>("voxel_grid", 1);
-  }
+  _voxel_pub = nh.advertise<sensor_msgs::PointCloud2>("voxel_grid", 1);
```
SteveMacenski: Please add the if statement back in, I dont want to create an unused publisher. This may show up on a rostopic list and confuse folks.

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -479,6 +487,45 @@ bool SpatioTemporalVoxelLayer::RemoveStaticObservations(void)
...
+  updateFlagIfChanged(_voxel_decay, config.voxel_decay);
+  updateFlagIfChanged(_decay_model, config.decay_model);
+  updateFlagIfChanged(_publish_voxels, config.publish_voxel_map);
+
+  _enabled = config.enabled;
+  _combination_method = config.combination_method;
+  _mark_threshold = config.mark_threshold;
+  _update_footprint_enabled = config.update_footprint_enabled;
+
+  if (update_grid){
```
SteveMacenski: Mhm, I feel like there's a better way.

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
veimox: When dynamic reconfigure server gets created, it calls the callback with the default configuration (with de values specified in the .cfg) so is common to handle this specific case separately. In other projects I use this as a initialization-like method too, but for this case I think is better to use the rosparams as (1) follows what you might have already and (2) it follows the move\_base way of configuring the system.

This variable could be removed if the server is initialized before all the variables that are got nowadays with rosparam but it might create race conditions. E.g. the code has a variable called "start\_working" this will get initialized 3 times: 
1. in the constructor of the object
2. in the dynamic reconfigure server
3. in rosparam

If this variable is used in another thread to do something this might create problems depending on where the thread is created. I put this variable in case it wanted to be used for initializing something, but it can be removed if the server is initialized before the rosparams.

src/spatio_temporal_voxel_layer.cpp
veimox: In order to create the publisher on the dynamic reconfigure callback, the node handler is needed. Either this object is created in the callback it self and then thrown away or the class has an instance of the node hadler that other methods can use. This instance can be either the object or a pointer to the object.

It is  a pointer because in the constructor is not meant to initialize anything in costmap\_2d and I think is a bit of waste create a full Node Handler to override it later on. All variables that are not meant to be initialized in the constructor but that need to be private members of the class I tend to create pointers so they can, later on, be filled. 

Ideally we would be utilizing the \_nh more than the nh and frieds, but as a prototype i didnt want to touch that part of the code until further discussed.

So far, I see three options:
1. NH in the callback
2. NH in class as object
3. NH in class as pointer

To me the cleanest is the third option

veimox: Oh! I have just realized that I didnt use it a all! Well, the comment still makes sense :stuck_out_tongue:

veimox:
Hi!

I completely forgot this PR, sorry :disappointed: . I can consider this PR dead. In the company I work for we decided to go for another approach and thus, I stopped working on this. 

You can continue this PR if needed and you can contact me if needed :smile: 

@SteveMacenski I let you close the PR if needed.

SteveMacenski:
@veimox Ok, no problem thanks for getting back to us :) 

Feel free to ping me on my email in the package.xml if there were particular issues at hand I could help resolve.

PR #87, opened 2018-11-08
CMakeLists.txt fix for ROS Melodic's catkin

discord-intech:
Under certain configurations, CMAKE_SOURCE_DIR doesn't point to the project directory.
PROJECT_SOURCE_DIR always point to the directory of the last project() command.


SteveMacenski:
Nice catch - thanks for the PR!

PR #91, opened 2018-11-19
Patch persistent voxel decay model returning the wrong duration

tasilb:
Testing this right now


tasilb:
Results look good

SteveMacenski:
@tasilb test and merge as you're ready

PR #93, opened 2018-11-20
implementing abstract class for configurable frustum models

SteveMacenski:
@nickvaras I implemented the abstract class so you can add your frustum model to the package.

As you can see you need to do a couple things:
* Add `THREE_D_LIDAR` OR `VLP16_OR_SIMILAR` to the `ModelType` enum
* implement your class as a derived class of the frustum model
* add an `if else` statement in the iterator loop for processing clearing measurements
* set the `model_type` parameter to 0, 1, etc for what model you like

This PR requires
- [x] testing to make sure nothing broke
- [x] validation of no memory leaks - look at in morning if still running

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,7 +82,7 @@ add_library(${PROJECT_NAME}
             src/spatio_temporal_voxel_layer.cpp
             src/spatio_temporal_voxel_grid.cpp
             src/measurement_buffer.cpp
-            src/frustum.cpp
+            src/frustum_models/depth_camera_frustum.cpp
             src/vdb2pc.cpp
 )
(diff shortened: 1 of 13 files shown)


nickvaras:
My move_base process keeps crashing when running this, but runs ok with the kinetic-devel branch... anyone getting the same results?

SteveMacenski:
@nickvaras I've done no testing here yet, I just did this last night at home. Will address some time this week. If you branch off of this branch so you have the abstract class you can get started regardless. When you want to merge,we can just rebase to kinetic-devel.

move_base crashing isnt good... probably some pointer issue - like I said - addressing some time this week. I'm swamped at work so dont want on this

PR #98, opened 2018-12-05
Dynamic reconfigure proposal continuing

MarcelSoler:
Taken from #86


SteveMacenski:
You have a couple of merge conflicts, can you resolve?

CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 2.8.3)
 project(spatio_temporal_voxel_layer)
-add_compile_options(-std=c++17)
+
+add_definitions (-std=c++14)
```
SteveMacenski: Can we bring this back to 17?

cfg/SpatioTemporalVoxelLayer.cfg
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python
```
SteveMacenski: Did you add everything that's a "layer" parameter (vs. an observation source parameter) and default to the coded defaults?

include/spatio_temporal_voxel_layer/measurement_buffer.hpp
```diff
@@ -101,6 +102,10 @@ class MeasurementBuffer
   // Get measurements from the buffer
   void GetReadings(std::vector<observation::MeasurementReading>& observations);
 
+  // enabler setter getter
+  bool isEnabled() const;
```
SteveMacenski: We have this repo set up so functions are all Caps, can you change to `IsEnabled` and `SetEnabled`? also, while you're at it, a `const bool&` I know its a small thing but its for consistency. Also add `(void)` in `IsEnabled(void)`.

All small things

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
```diff
@@ -76,6 +79,8 @@ namespace spatio_temporal_voxel_layer
 // conveniences for line lengths
 typedef std::vector<boost::shared_ptr<message_filters::SubscriberBase> >::iterator observation_subscribers_iter;
 typedef std::vector<boost::shared_ptr<buffer::MeasurementBuffer> >::iterator observation_buffers_iter;
+typedef spatio_temporal_voxel_layer::SpatioTemporalVoxelLayerConfig dynamicReconfigureType;
+typedef  dynamic_reconfigure::Server<dynamicReconfigureType> dynamicReconfigureServerType;
```
SteveMacenski: Looks like an extra space?

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
```diff
@@ -126,25 +131,36 @@ class SpatioTemporalVoxelLayer : public costmap_2d::CostmapLayer
   bool AddStaticObservations(const observation::MeasurementReading& obs);
   bool RemoveStaticObservations(void);
 
+  // Dynamic reconfigure
+  void DynamicReconfigureCallback(dynamicReconfigureType &config, uint32_t level);
+  dynamicReconfigureServerType* _dynamic_reconfigure_server;
+
+  void BufferEnablerCallback( std_srvs::SetBool::Request & request,    \
```
SteveMacenski: Oh, those nice aligning `\`, makes me OCD happy :)

src/measurement_buffer.cpp
```diff
@@ -275,6 +276,21 @@ bool MeasurementBuffer::UpdatedAtExpectedRate(void) const
...
+
+/*****************************************************************************/
+bool MeasurementBuffer::isEnabled() const
+/*****************************************************************************/
+{
+  return _enabled;
+}
+
+/*****************************************************************************/
+void MeasurementBuffer::setEnabled(const bool enabled)
```
SteveMacenski: Do you need a mutex here since setting could be done async?

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -62,15 +63,15 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
 
   // initialize parameters, grid, and sub/pubs
   ros::NodeHandle nh("~/" + name_), g_nh, prefix_nh;
+  _nh = &nh;
```
SteveMacenski: This is unused, can you remove _nh completely, I'm not sure why it was added in the other PR

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -114,16 +115,17 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
 
   if (_publish_voxels)
   {
-    _voxel_pub = nh.advertise<sensor_msgs::PointCloud2>("voxel_grid", 1);
+     _voxel_pub = nh.advertise<sensor_msgs::PointCloud2>("voxel_grid", 1);
```
SteveMacenski: added space?

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -192,7 +195,7 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
         obstacle_range, *tf_, _global_frame,                             \
         sensor_frame, transform_tolerance, min_z, max_z, vFOV,           \
         hFOV, decay_acceleration, marking, clearing, _voxel_size,        \
-        voxel_filter, clear_after_reading)));
+        voxel_filter, enabled,  clear_after_reading)));
```
SteveMacenski: 1 too many spaces

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -479,6 +523,42 @@ bool SpatioTemporalVoxelLayer::RemoveStaticObservations(void)
   }
 }
 
+/*****************************************************************************/
+void SpatioTemporalVoxelLayer::DynamicReconfigureCallback( \
```
SteveMacenski: This might be my ignorance of dynamic reconfigure, but does it send params even if they don't change? ie to fill out the `config` message for variable `x` does it send `y` as the default value even though you didnt want to change it? Otherwise some of these variables set below would be consistently overwritten if you just wanted to enable/disable through this mechanism

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -334,6 +349,35 @@ void SpatioTemporalVoxelLayer::PointCloud2Callback( \
...
+                std_srvs::SetBool::Response & response, \
+                boost::shared_ptr<buffer::MeasurementBuffer>& buffer, \
+                boost::shared_ptr<message_filters::SubscriberBase>& subcriber)
+/*****************************************************************************/
+{
+  buffer->Lock();
+  if( buffer->isEnabled() != request.data ) //updated
+  {
+    buffer->setEnabled(request.data);
+    if(request.data == true)
```
SteveMacenski: `if(request.data)`, the true is redundant

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -334,6 +349,35 @@ void SpatioTemporalVoxelLayer::PointCloud2Callback( \
...
+    {
+      subcriber->subscribe();
+    }
+    else
+    {
+      if (subcriber != NULL)
```
SteveMacenski: `if (subscriber)`, will return false if its NULL

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -249,6 +252,11 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
           boost::bind(&SpatioTemporalVoxelLayer::PointCloud2Callback, this, _1, \
                                                    _observation_buffers.back()));
 
+      boost::shared_ptr <ros::ServiceServer> server(new ros::ServiceServer);
```
SteveMacenski: Please test this and make sure it does what you expect it to. Can you add an example in the readme file of how to use this to turn on/off the observation sources (and what topics/service names they come up under)?

Generally speaking, it looks clean, good work

MarcelSoler:
Hi, It' still wip, on Monday I'll continue, I just pushed it to don't lose any local changes .

Quick thoughts: 
-  I've realised is that in dynamic reconfigure you cannot have dynamic elements. Hence you cannot have a enabled button for each layer.
-  What I'm doing is creating a service for each observation source. The topic will be it's own observation_name. 
This way you can send a bool service each observation to enable or disable them.  Similar to the the enable button of the whole layer.
When I have successfully tested it I'll add it to the READme.

src/spatio_temporal_voxel_layer.cpp
MarcelSoler: The service can be true or false. To enable /disable the observation.  With that if I check  if you want to activate or deactivate the observation.

cfg/SpatioTemporalVoxelLayer.cfg
MarcelSoler: not yet I haven't review the reconfigure part yet.

CMakeLists.txt
MarcelSoler: We can, but I don't see the reason since there aren't any c++17 functions in use right now . Anyway I'll change it if you think it's better.

SteveMacenski:
Ok,
The readme element would be pretty important for people to utilize it and make sure its own to exist since it really is a good idea. 

Thanks for the contribution!

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
```diff
@@ -126,21 +131,32 @@ class SpatioTemporalVoxelLayer : public costmap_2d::CostmapLayer
...
+                              boost::shared_ptr<message_filters::SubscriberBase>& subcriber);
+
   laser_geometry::LaserProjection                                  _laser_projector;
   std::vector<boost::shared_ptr<message_filters::SubscriberBase> > _observation_subscribers;
   std::vector<boost::shared_ptr<tf::MessageFilterBase> >           _observation_notifiers;
   std::vector<boost::shared_ptr<buffer::MeasurementBuffer> >       _observation_buffers;
   std::vector<boost::shared_ptr<buffer::MeasurementBuffer> >       _marking_buffers;
   std::vector<boost::shared_ptr<buffer::MeasurementBuffer> >       _clearing_buffers;
+  //std::vector<boost::shared_ptr<ros::ServiceServer> >              _buffer_enabler_servers;
+  std::vector<ros::ServiceServer*>                                 _buffer_enabler_servers;
```
SteveMacenski: If you're switching from boost pointers, remember to delete them in the destructor

src/spatio_temporal_voxel_layer.cpp
MarcelSoler: For what I know:
Dynamic reconfigure at the init first loads the dynamic reconfigure.cfg then its values are overwritten by ros_params (commom_params.yaml). 
When in run time some params are changed with the dynamic reconfigure, all of them are updated. However the value is updated with the dynamic reconfigured config which is the one obtained from ros_params so I wont be any problems. I'll test it anyway.
I hope I explained clearly...

cfg/SpatioTemporalVoxelLayer.cfg
MarcelSoler: Problems encountered:
-  It's not possible to change in real-time any parameters belonging to observation sources though reconfigure... Like:

max_obstacle_height
obstacle_range
...

I think the only worth adding are:
transform_tolerance
mapping_mode 
map_save_duration

And with the last two I have some doubts, if they have any affect in real time or are only called in the init(). Can you clarify please?

src/spatio_temporal_voxel_layer.cpp
SteveMacenski: Ok sounds good - given that the callback runs after the ROS params are evaluated (in the first iteration) I would think it would overwrite, unless we did the dynamic reconfigure callback registering **before** we call ros params (we might want to just move it so there's no chance). 

Thanks for the explanation

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
```diff
@@ -147,7 +147,7 @@ class SpatioTemporalVoxelLayer : public costmap_2d::CostmapLayer
   std::vector<boost::shared_ptr<buffer::MeasurementBuffer> >       _marking_buffers;
   std::vector<boost::shared_ptr<buffer::MeasurementBuffer> >       _clearing_buffers;
   //std::vector<boost::shared_ptr<ros::ServiceServer> >              _buffer_enabler_servers;
-  std::vector<ros::ServiceServer*>                                 _buffer_enabler_servers;
+  std::vector<ros::ServiceServer>                                 _buffer_enabler_servers;
```
SteveMacenski: Good call, no real reason to need a pointer on this guy

MarcelSoler:
I tested it today and it works! Tomorrow I'll play a bit more and edit the read-me.

SteveMacenski:
Awesome! It looks like you have some merge conflicts but I'm excited to see :)

src/spatio_temporal_voxel_layer.cpp
MarcelSoler: Tested worked as expected! params are not everrided.

src/spatio_temporal_voxel_layer.cpp
SteveMacenski: Sorry, what i mean is the line can be `if(request.data)` instead of `if(request.data == true)`

CMakeLists.txt
SteveMacenski: Please do change back, sorry its just a future proofing thing

cfg/SpatioTemporalVoxelLayer.cfg
SteveMacenski: Yes that makes sense, I'm looking to just have everything [here](https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/kinetic-devel/example/standard_indoor_environment_config.yaml#L18) and above - just the "layer" parameters, not the "observation" specific ones. We could have dynamic reconfigure actually reconfigure the buffer vectors as well but that's out of the scope of this PR and what you're trying to do

I think missing are
- mapping_mode
- map_save_duration

SteveMacenski:
@MarcelSoler I have a couple last little things in the review above and then its good to go.

I can look at your issue you're describing next week - I'm a little slammed right now. That is strange, whats your ideal outcome? They both clear, neither clear? (are they fully clearing or just clearing out that sensor?) Also, do you have your local decay params set such that they wouldnt clear as fast so during that time it would stick around?

SteveMacenski:
Clearing moving people is a function of the voxel acceleration, model, voxel decay, you just need to play around with those values until you get something you're happy with.They're also function of the update cycle frequency for the acceleration so its a thing you should first find your update rate, calculate what kind of global decay you'd like, then tune the accleration to match. It does take some time but its the price we pay for such lower CPU draw.

I dont think the local costmap problem you describe is a function that you have both marking/clearing in the same source - but it should be an easy test for you to make sure. Let me know. 

 What are your other local costmap settings (rolling, freq, etc). Have you tried just copy pasting the same params from the global costmap into your local and changing them 1 by 1 until you isolate the culprit? Or is it generally just all params for the local that make this happen.

MarcelSoler:
I've been doing a few tests trying to solve the two problems:
1. That  local map doesn't clear obstacles decaying them:
    -  If i do it with the kinetic_devel branch with the same move_base parameters it works as expected. So i don't thing the culprit is some costmap param, I think I created it with some commit ...

2.  Clearing function: I tried separating again clearing marking observations and the result is the same it doesn't clear.
I also tried tuning, playing with several voxel_decay and decay_acceleration not luck for now... 
My sensor update rate is 30Hz and I think the FOV and HOV are more or less adjusted.

I'll keep trying! any thoughts will be appreciated.

SteveMacenski:
I'm glad you figured it out -- so its just tuning now is your issue, correct? Can you send me a video of what you have so far and we can go from there? More than willing to help given all the hard work you've done here.

Anything else holding up this PR? Or just me messing around with it to make sure we're all good

MarcelSoler:
Nope , good to go from my side. It would be nice if you can test it out. I've done a few test with our robot and works fine, the change cameras of cameras is real quick. However I would like to someone else to test it out.
Maybe we can merge it and solve the issues as they come ;)

MarcelSoler:
Ill add the final touches: I want to disable the services if the user doesn't want them, I'll add a bool to do so.
Also being able to start with the sensor disabled. I'll test it right away

SteveMacenski:
A few questions on the recent changes: why add a parameter to disable the services? I think it's totally something that can just be there and if you use them, you use them. Disabling them adds complexity and less flexibility and I don't see a downside to advertising an extra service that's just unused. 

I also think you did the right thing to have `buffer->ResetLastUpdatedTime();`, can you add it back in? Why did you remove it?



I'll test this tomorrow.

MarcelSoler:
> A few questions on the recent changes: why add a parameter to disable the services? I think it's totally something that can just be there and if you use them, you use them. Disabling them adds complexity and less flexibility and I don't see a downside to advertising an extra service that's just unused.
> 
> I also think you did the right thing to have `buffer->ResetLastUpdatedTime();`, can you add it back in? Why did you remove it?

ok! I added the param for testing and I thought it could be useful.  However it's true that adds complexity. I'll roll it back to what it used to be.

MarcelSoler:
Done! ready to go.

SteveMacenski:
@MarcelSoler I find that dynamic reconfigure crashes once every 10-20 changes, I think its due to the fact beforehand all of the interactions with `_voxel_grid` were synchronous, and now we're deleting and remaking it without a mutex.

I would recommend adding a mutex for `_voxel_grid_lock` and any function/call to `_voxel_grid` is locked down. OpenVDB isn't meant to have asynchronous changes to it. Async gets are ok, but async sets/deletes can cause this

SteveMacenski:
Otherwise, the toggling services work exactly as I would expect, great job! Dynamic reconfigure will certainly make many people's lives easier. 

Just these 2 things and I can just merge it. I have some cleanup I'll merge in afterwards but nothing I'll make you do (just my OCD flairing up :) ).

SteveMacenski:
Ok, I'll look when I get back to work on Jan 1. But assuming that works I'll give you the approval to merge.

SteveMacenski:
Crashing makes people unhappy which then makes the number of emails I get a week rise 20% :)

MarcelSoler:
Tested ok! from my side I played changing parameters with reconfigure like 10 minutes and it didn't crash.

SteveMacenski:
Awesome, will test today and give the approval for you to merge (want you to get on the contributor scoreboard!)

SteveMacenski:
Ok tested more @MarcelSoler, it definitely crashes less now. However occasionally when the parameters that require a new `_voxel_grid` to be made it crashes. I suspect it has to do with the mutex, that mutex is meant for the data structure underlying the grid, not the grid itself. So when someone in the normal update cycle tries to use one of the auxiliary methods that doesn't directly call the underlying data structure (GetFlattenedCostmap, etc) it will crash since the `_voxel_grid` doesnt exist.

Also you delete the object (and then the mutex goes with it) then Unlock() after the deletion where that new lock is still unlocked.

I would recommend making the mutex in the layer `_voxel_grid_lock` which is the mutex for accessing the grid itself, that way it persists between different `_voxel_grid`s and locks the full grid and not the internal data structure.

SteveMacenski:
You can test this by just toggling `track_unknown_space` for a few moments/minutes since that will make a new `_voxel_grid`.

MarcelSoler:
With the last commit the multi threading problem has been corrected (at least i think so). The grid is "protected". I've encountered one last error :

When in the SpatioTemporalVoxelLayer::onInitialize(void)  publish_voxel_map is false :
```
  if (_publish_voxels)
  {
    _voxel_pub = nh.advertise<sensor_msgs::PointCloud2>("voxel_grid", 1);
  }
```
The publisher is not advertised. If afterwards in the reconfigure it's set to true, tries to publish with a non existing publisher.
```
[FATAL] [/navigation/move_base]: ASSERTION FAILED
	file = /opt/ros/kinetic/include/ros/publisher.h
	line = 102
	cond = false
	message = 
[FATAL] [/navigation/move_base]: Call to publish() on an invalid Publisher
[FATAL] [/navigation/move_base]: 
```
One option will be to advertise it always. I'll to test it out in a new commit.

MarcelSoler:
@SteveMacenski  tested ok! all correct from my side.

SteveMacenski:
Awesome, this looks good to me as well. I have some cleanup I'll PR this afternoon but I won't make you do it. 

I added you as a collaborator so you can merge the PR. Approved!

MarcelSoler:
Thanks for your guidance!!!

PR #108, opened 2019-01-18
WIS STVL Running on melodic

Patrick-Lascombe:
Adding an intermediate transform to use pcl_ros::transformPointCloud.
There might be some useless dependencies in package.xml and CMakeLists.txt atm.

Next step : 
Correcting this error : 
`TF Exception for sensor frame: , cloud frame: camera_nav_depth_optical_frame, Lookup would require extrapolation into the past.  Requested time 6.145000000 but the earliest data is at time 7.210000000, when looking up transform from frame [camera_nav_depth_optical_frame] to frame [odom]`.

After a while this error is affecting all the tfs used by the costmap.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,6 @@ find_package(catkin REQUIRED COMPONENTS
   tf2
   tf2_geometry_msgs
   tf2_sensor_msgs
-  tf2_tools
   visualization_msgs
   message_generation
 )
@@ -75,8 +74,6 @@ catkin_package(
         tf2
         tf2_geometry_msgs
         tf2_conversions
-        tf2_bullet
-        rostime
         visualization_msgs
     DEPENDS
         TBB
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,10 +15,7 @@ find_package(catkin REQUIRED COMPONENTS
   pcl_conversions
   pcl_ros
   roscpp
-  tf
-  tf2
-  tf2_geometry_msgs
-  tf2_sensor_msgs
+  tf2_ros
   visualization_msgs
   message_generation
 )
@@ -71,9 +68,7 @@ catkin_package(
         sensor_msgs
         std_msgs
         costmap_2d
-        tf2
-        tf2_geometry_msgs
-        tf2_conversions
+        tf2_ros
         visualization_msgs
     DEPENDS
         TBB
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,9 +18,7 @@ find_package(catkin REQUIRED COMPONENTS
   tf
   tf2
   tf2_geometry_msgs
-  tf2_ros
   tf2_sensor_msgs
-  tf2_msgs
   tf2_tools
   visualization_msgs
   message_generation
@@ -35,7 +33,7 @@ generate_messages(
    std_msgs
 )
 
-find_package(PCL REQUIRED)#1.7.2 EXACT)
+find_package(PCL REQUIRED)
 find_package(OpenVDB REQUIRED)
 find_package(TBB REQUIRED)
 remove_definitions(-DDISABLE_LIBUSB-1.0)
@@ -75,8 +73,6 @@ catkin_package(
         std_msgs
         costmap_2d
         tf2
-        tf2_ros
-        tf2_msgs
         tf2_geometry_msgs
         tf2_conversions
         tf2_bullet
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,10 +15,13 @@ find_package(catkin REQUIRED COMPONENTS
   pcl_conversions
   pcl_ros
   roscpp
+  tf
   tf2
   tf2_geometry_msgs
   tf2_ros
   tf2_sensor_msgs
+  tf2_msgs
+  tf2_tools
   visualization_msgs
   message_generation
 )
@@ -32,7 +35,7 @@ generate_messages(
    std_msgs
 )
 
-find_package(PCL REQUIRED)
+find_package(PCL REQUIRED)#1.7.2 EXACT)
 find_package(OpenVDB REQUIRED)
 find_package(TBB REQUIRED)
 remove_definitions(-DDISABLE_LIBUSB-1.0)
@@ -72,6 +75,12 @@ catkin_package(
         std_msgs
         costmap_2d
         tf2
+        tf2_ros
+        tf2_msgs
+        tf2_geometry_msgs
+        tf2_conversions
+        tf2_bullet
+        rostime
         visualization_msgs
     DEPENDS
         TBB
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -15,10 +15,13 @@ find_package(catkin REQUIRED COMPONENTS
   pcl_conversions
   pcl_ros
   roscpp
+  tf
   tf2
   tf2_geometry_msgs
   tf2_ros
   tf2_sensor_msgs
+  tf2_msgs
```
SteveMacenski: I don't see these being used anywhere - are they needed for this package?

CMakeLists.txt
```diff
@@ -32,7 +35,7 @@ generate_messages(
    std_msgs
 )
 
-find_package(PCL REQUIRED)
+find_package(PCL REQUIRED)#1.7.2 EXACT)
```
SteveMacenski: You can remove the EXACT statement :)

CMakeLists.txt
```diff
@@ -72,6 +75,12 @@ catkin_package(
         std_msgs
         costmap_2d
         tf2
+        tf2_ros
```
SteveMacenski: There's lot of packages added here - are you sure they're all needed? tf2_ros and tf2_geometry_msgs seems fine - many of these arent in find_package either (conversions, bullet, rostime, etc, I dont think any of these are necessary)

package.xml
```diff
@@ -37,11 +45,20 @@
   <run_depend>std_msgs</run_depend>
   <run_depend>roscpp</run_depend>
   <run_depend>tf2</run_depend>
+  <run_depend>tf2_ros</run_depend>
+  <run_depend>tf2_msgs</run_depend>
```
SteveMacenski: Same comments in CMakeLists file about the need for some of these.

include/spatio_temporal_voxel_layer/measurement_buffer.hpp
```diff
@@ -53,9 +53,13 @@
 #include <ros/time.h>
 // TF
 #include "tf2_ros/transform_listener.h"
+#include "tf/transform_listener.h"
```
SteveMacenski: I'm not sure why TF1 is needed, can you explain or show me where something from this header is used in the changes?

include/spatio_temporal_voxel_layer/measurement_buffer.hpp
```diff
@@ -126,6 +130,7 @@ class MeasurementBuffer
   double _min_z, _max_z, _vertical_fov, _horizontal_fov, _decay_acceleration;
   double _voxel_size;
   bool _marking, _clearing, _voxel_filter, _clear_buffer_after_reading;
+  geometry_msgs::TransformStamped tf_stamped;
```
SteveMacenski: This doesnt need to be a class variable, you can just instantiate it in the use below as 
```
geometry_msgs::TransformStamped tf_stamped = _buffer.lookupTransform(_global_frame, cloud.header.frame_id.c_str(), ros::Time(0));

```

src/measurement_buffer.cpp
```diff
@@ -152,7 +152,21 @@ void MeasurementBuffer::BufferPCLCloud(const \
...
+    tf::Quaternion q(tf_stamped.transform.rotation.x,
+                     tf_stamped.transform.rotation.y,
+                     tf_stamped.transform.rotation.z,
+                     tf_stamped.transform.rotation.w);
+
+    tf::Vector3 t(tf_stamped.transform.translation.x,
+                  tf_stamped.transform.translation.y,
+                  tf_stamped.transform.translation.z);
+
+    tf::Transform transform(q, t);
```
SteveMacenski: This can be streamlined using TF's built in conversion options: http://docs.ros.org/melodic/api/tf/html/c++/transform__datatypes_8h.html

You can use ```tf::transformMsgToTF (const geometry_msgs::Transform &msg, Transform &bt)``` to directly go from geometry_msgs::Transform to tf::Transform in 1 line. Would change this section to:

```
tf::Transform tf_transform;
tf_stamped = _buffer.lookupTransform(_global_frame, cloud.header.frame_id.c_str(), ros::Time(0));
tf::transformMsgToTF(tf_stamped, tf_transform)
```

SteveMacenski:
All the changes look good, I'm still curious on the deps for `ros_time` `tf2_msgs` `tf2_bullet` and `tf2_tools`

Patrick-Lascombe:
I'll look them up, see which of them is used and delete the others

CMakeLists.txt
```diff
@@ -15,9 +15,9 @@ find_package(catkin REQUIRED COMPONENTS
   pcl_conversions
   pcl_ros
   roscpp
+  tf
   tf2
   tf2_geometry_msgs
-  tf2_ros
```
SteveMacenski: You probably want this to be included in here + package xml if you removed it

SteveMacenski:
Personally, I look at the headers I added or removed and add or remove those from the CMakeLists + package xml. It looks like you added some TF2_ROS headers so that should be the only thing you need to add (and should remove anything TF since that's not used in melodic) 

Then try compiling it in a clean environment (Docker, new machine, etc should work) off of the osrf/melodic-desktop image. I don't try to go too crazy in dependency management as long as there's not a bunch of stuff being added that don't appear to be used

src/measurement_buffer.cpp
```diff
@@ -152,7 +152,14 @@ void MeasurementBuffer::BufferPCLCloud(const \
 
     point_cloud_ptr cld_global(new pcl::PointCloud<pcl::PointXYZ>);
 
-    _buffer.transform(cloud, *cld_global, _global_frame);
+    // Create a tf::Transform using the tfStamped for _buffer
+    geometry_msgs::TransformStamped tf_stamped = _buffer.lookupTransform(_global_frame, cloud.header.frame_id.c_str(), ros::Time(0));
+    geometry_msgs::TransformStamped tf_stamped = _buffer.lookupTransform(_global_frame, cloud.header.frame_id.c_str(), ros::Time::now(), ros::Duration(5.0));
```
SteveMacenski: I assume you only need one of these 2 `lookupTransform` calls? Once you resolve that, the rest of this looks good (though you need to include the header `#include <geometry_msgs/TransformStamped.h>` in `measurement_buffer.hpp`).

src/measurement_buffer.cpp
```diff
@@ -152,7 +152,13 @@ void MeasurementBuffer::BufferPCLCloud(const \
 
     point_cloud_ptr cld_global(new pcl::PointCloud<pcl::PointXYZ>);
 
-    _buffer.transform(cloud, *cld_global, _global_frame);
+    // Create a tf::Transform using the tfStamped for _buffer
+    geometry_msgs::TransformStamped tf_stamped = _buffer.lookupTransform(_global_frame, cloud.header.frame_id.c_str(), ros::Time::now(), ros::Duration(5.0));
```
SteveMacenski: Very last thing: you have the timeout to 5 seconds, that's a huge amount, it should be something like 0.1 seconds that sounds more reasonable. 

Can you give me the :+1: that this is all working now? Will merge

Patrick-Lascombe: It is working !
Here are 2 videos of the layer working on our robot : 
- https://youtu.be/Ziaj99DUJgo
- https://youtu.be/71uVkSqQ3G8

The slow update is most likely due to our platform limitation

SteveMacenski:
Approved! I added you as a collaborator so you can squash merge your own PR! Thanks for the amazing help!

Patrick-Lascombe:
No problem at all, thank you a lot for providing a tool like this one ! I'll keep in touch if we find other bug or improvement with melodic

Issue #117, opened 2019-02-26
[melodic] pcl::fromPCLPointCloud2 performance issue


doisyg:
As pointed in https://github.com/SteveMacenski/spatio_temporal_voxel_layer/issues/110, under Melodic, we observe a serious performance issue that we traced to the **PCL** fonction  `pcl::fromPCLPointCloud2`. Possibly caused by the change of PCL version between Kinetic and Melodic (1.7.x to 1.8.x) but not investigated further.
With Melodic, on a i5 computer move_base completely overload a core the CPU with a 30Hz 320x240 poincloud (about 74MB/s). Whereas with kinetic and the same config, move_base use about 40% of a core.

To avoid using `pcl::fromPCLPointCloud2`, I did a test by modifying the Pointcloud2 Callback:
https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/faaf1498649e666bbd6901a98ae135c1d51a5536/src/spatio_temporal_voxel_layer.cpp#L326-L328
to get the message directly in the `pcl::PointCloud<pcl::PointXYZ>` format, and let roscpp take care of the converstion (see http://wiki.ros.org/pcl_ros#ROS_C.2B-.2B-_interface). 
```
void SpatioTemporalVoxelLayer::PointCloud2Callback( \
                const pcl::PointCloud<pcl::PointXYZ>::ConstPtr& message, \
                const boost::shared_ptr<buffer::MeasurementBuffer>& buffer)
```
The performance is better and makes STVL usable with melodic but there is still a decrease of performance compared to Kinetic (65% CPU). I can do a quick PR if someone is interested.

SteveMacenski:
@doisyg have you looked at the migration for TF -> TF2 (also just melodic) in other projects? Costmap 2d comes to mind and this is the way their buffer handles it (see alot of similarities ;) ) 

https://github.com/moriarty/navigation/blob/57399e086e8bd4174f6f3f8c938bc8cefe69c5b0/costmap_2d/src/observation_buffer.cpp#L108

and its PR https://github.com/ros-planning/navigation/pull/755/files

I wouldnt recommend necessarily working with the pcl::pointcloud as the raw input since that's not the API anyone else uses and depth cameras themselves by an REP standard requires sensor_msgs::PointCloud2.

You sure its not just an issue with how its being done? I don't see any melodic reports of their buffers getting heavier with melodic

SteveMacenski:
[...]

PR #120, opened 2019-03-05
Rebase STVL on PointCloud2 structure

doisyg:
First working attempt to rebase STVL on PointCloud2 to overcome performance issue. See discussion https://github.com/SteveMacenski/spatio_temporal_voxel_layer/issues/117

### DIFF
--- a/include/spatio_temporal_voxel_layer/measurement_buffer.hpp
+++ b/include/spatio_temporal_voxel_layer/measurement_buffer.hpp
@@ -60,6 +60,7 @@
 #include <sensor_msgs/PointCloud2.h>
 #include <geometry_msgs/Quaternion.h>
 #include <geometry_msgs/PoseStamped.h>
+#include <geometry_msgs/TransformStamped.h>
 // Mutex
 #include <boost/thread.hpp>
 
--- a/include/spatio_temporal_voxel_layer/measurement_buffer.hpp
+++ b/include/spatio_temporal_voxel_layer/measurement_buffer.hpp
@@ -43,7 +43,9 @@
 #include <spatio_temporal_voxel_layer/measurement_reading.h>
 // PCL
 #include <pcl_ros/transforms.h>
-#include <pcl/filters/approximate_voxel_grid.h>
+#include <pcl/filters/voxel_grid.h>
+#include <pcl_conversions/pcl_conversions.h>
+#include <pcl/filters/passthrough.h>
 // STL
 #include <vector>
 #include <list>
@@ -60,14 +62,13 @@
 #include <geometry_msgs/PoseStamped.h>
 // Mutex
 #include <boost/thread.hpp>
-#include <geometry_msgs/TransformStamped.h>
 
 namespace buffer
 {
 
 // conveniences for line lengths
 typedef std::list<observation::MeasurementReading>::iterator readings_iter;
-typedef pcl::PointCloud<pcl::PointXYZ>::Ptr point_cloud_ptr;
+typedef sensor_msgs::PointCloud2::Ptr point_cloud_ptr;
 
 // Measurement buffer
 class MeasurementBuffer
@@ -98,7 +99,6 @@ public:
 
   // Buffers for different types of pointclouds
   void BufferROSCloud(const sensor_msgs::PointCloud2& cloud);
-  void BufferPCLCloud(const pcl::PointCloud<pcl::PointXYZ>& cloud);
 
   // Get measurements from the buffer
   void GetReadings(std::vector<observation::MeasurementReading>& observations);
--- a/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
+++ b/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
@@ -51,9 +51,12 @@
 #include <ctime>
 #include <iostream>
 #include <utility>
+#include <list>
 // msgs
 #include <sensor_msgs/PointCloud2.h>
+#include <sensor_msgs/point_cloud2_iterator.h>
 #include <visualization_msgs/Marker.h>
+#include <geometry_msgs/Point.h>
 // TBB
 #include <tbb/parallel_do.h>
 // OpenVDB
@@ -116,7 +119,7 @@ public:
   void ClearFrustums(const std::vector<observation::MeasurementReading>& clearing_observations);
 
   // Get the pointcloud of the underlying occupancy grid
-  void GetOccupancyPointCloud(pcl::PointCloud<pcl::PointXYZ>::Ptr& pc);
+  void GetOccupancyPointCloud(sensor_msgs::PointCloud2::Ptr& pc2);
   std::unordered_map<occupany_cell, uint>* GetFlattenedCostmap();
 
   // Clear the grid
@@ -153,7 +156,7 @@ protected:
   int                             _decay_model;
   double                          _background_value, _voxel_size, _voxel_decay;
   bool                            _pub_voxels;
-  pcl::PointCloud<pcl::PointXYZ>::Ptr     _pc;
+  std::list<geometry_msgs::Point32>*   _grid_points;
   std::unordered_map<occupany_cell, uint>* _cost_map;
   boost::mutex                            _grid_lock;
 };
--- a/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
+++ b/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
@@ -51,7 +51,7 @@
 #include <ctime>
 #include <iostream>
 #include <utility>
-#include <list>
+#include <vector>
 // msgs
 #include <sensor_msgs/PointCloud2.h>
 #include <sensor_msgs/point_cloud2_iterator.h>
@@ -156,7 +156,7 @@ protected:
   int                             _decay_model;
   double                          _background_value, _voxel_size, _voxel_decay;
   bool                            _pub_voxels;
-  std::list<geometry_msgs::Point32>*   _grid_points;
+  std::vector<geometry_msgs::Point32>*   _grid_points;
   std::unordered_map<occupany_cell, uint>* _cost_map;
   boost::mutex                            _grid_lock;
 };
--- a/src/measurement_buffer.cpp
+++ b/src/measurement_buffer.cpp
@@ -85,7 +85,7 @@ MeasurementBuffer::~MeasurementBuffer(void)
 void MeasurementBuffer::BufferROSCloud(const sensor_msgs::PointCloud2& cloud)
 /*****************************************************************************/
 {
-  try
+  tryspatio_temporal_voxel_layer
   {
     pcl::PCLPointCloud2 pcl_pc2;
     pcl_conversions::toPCL(cloud, pcl_pc2);
@@ -151,14 +151,12 @@ void MeasurementBuffer::BufferPCLCloud(const \
     }
 
     point_cloud_ptr cld_global(new pcl::PointCloud<pcl::PointXYZ>);
-
-    // Create a tf::Transform using the tfStamped for _buffer
-    geometry_msgs::TransformStamped tf_stamped = _buffer.lookupTransform(_global_frame, cloud.header.frame_id.c_str(), ros::Time::now(), ros::Duration(5.0));
-
-    tf::Transform tf_transform;
+    geometry_msgs::TransformStamped tf_stamped = _buffer.lookupTransform(_global_frame, cloud.header.frame_id.c_str(), pcl_conversions::fromPCL(cloud.header).stamp);
+    tf::StampedTransform tf_transform;
     tf::transformMsgToTF(tf_stamped.transform, tf_transform);
+    tf_transform.frame_id_ = _global_frame;
+    tf_transform.stamp_ = tf_stamped.header.stamp;
     pcl_ros::transformPointCloud(cloud, *cld_global, tf_transform);
-
     cld_global->header.stamp = cloud.header.stamp;
 
     // if user wants to use a voxel filter
--- a/src/measurement_buffer.cpp
+++ b/src/measurement_buffer.cpp
@@ -131,8 +131,8 @@ void MeasurementBuffer::BufferROSCloud(const sensor_msgs::PointCloud2& cloud)
 
     // transform the cloud in the global frame
     point_cloud_ptr cld_global(new sensor_msgs::PointCloud2());
-		geometry_msgs::TransformStamped tf_stamped = _buffer.lookupTransform( \
-		             _global_frame, cloud.header.frame_id, cloud.header.stamp);
+    geometry_msgs::TransformStamped tf_stamped = _buffer.lookupTransform( \
+                 _global_frame, cloud.header.frame_id, cloud.header.stamp);
     tf2::doTransform (cloud, *cld_global, tf_stamped);
 
     pcl::PCLPointCloud2::Ptr cloud_pcl (new pcl::PCLPointCloud2 ());
@@ -140,33 +140,33 @@ void MeasurementBuffer::BufferROSCloud(const sensor_msgs::PointCloud2& cloud)
 
     pcl_conversions::toPCL(*cld_global, *cloud_pcl);
 
-		// remove points that are below or above our height restrictions, and
-		// in the same time, remove NaNs and if user wants to use it, combine with a
+    // remove points that are below or above our height restrictions, and
+    // in the same time, remove NaNs and if user wants to use it, combine with a
     if ( _voxel_filter )
     {
-			pcl::VoxelGrid<pcl::PCLPointCloud2> sor;
-			sor.setInputCloud (cloud_pcl);
-			sor.setFilterFieldName("z");
-			sor.setFilterLimits(_min_obstacle_height,_max_obstacle_height);
-			sor.setDownsampleAllData(false);
-			sor.setLeafSize ((float)_voxel_size,
-			                 (float)_voxel_size,
-			                 (float)_voxel_size);
-			sor.filter(*cloud_filtered);
+      pcl::VoxelGrid<pcl::PCLPointCloud2> sor;
+      sor.setInputCloud (cloud_pcl);
+      sor.setFilterFieldName("z");
+      sor.setFilterLimits(_min_obstacle_height,_max_obstacle_height);
+      sor.setDownsampleAllData(false);
+      sor.setLeafSize ((float)_voxel_size,
+                       (float)_voxel_size,
+                       (float)_voxel_size);
+      sor.filter(*cloud_filtered);
     }
     else
     {
-			pcl::PassThrough<pcl::PCLPointCloud2> pass_through_filter;
-			pass_through_filter.setInputCloud(cloud_pcl);
-			pass_through_filter.setKeepOrganized(false);
-			pass_through_filter.setFilterFieldName("z");
-			pass_through_filter.setFilterLimits( \
-			            _min_obstacle_height,_max_obstacle_height);
-			pass_through_filter.filter(*cloud_filtered);
+      pcl::PassThrough<pcl::PCLPointCloud2> pass_through_filter;
+      pass_through_filter.setInputCloud(cloud_pcl);
+      pass_through_filter.setKeepOrganized(false);
+      pass_through_filter.setFilterFieldName("z");
+      pass_through_filter.setFilterLimits( \
+                  _min_obstacle_height,_max_obstacle_height);
+      pass_through_filter.filter(*cloud_filtered);
     }
 
-		pcl_conversions::fromPCL(*cloud_filtered, *cld_global);
-		_observation_list.front()._cloud = cld_global;
+    pcl_conversions::fromPCL(*cloud_filtered, *cld_global);
+    _observation_list.front()._cloud = cld_global;
   }
   catch (tf::TransformException& ex)
   {
@@ -221,8 +221,7 @@ void MeasurementBuffer::RemoveStaleObservations(void)
   for (it = _observation_list.begin(); it != _observation_list.end(); ++it)
   {
     observation::MeasurementReading& obs = *it;
-    const ros::Duration time_diff = \
-            _last_updated - obs._cloud->header.stamp;
+    const ros::Duration time_diff = _last_updated - obs._cloud->header.stamp;
 
     if (time_diff > _observation_keep_time)
     {
--- a/src/measurement_buffer.cpp
+++ b/src/measurement_buffer.cpp
@@ -85,7 +85,7 @@ MeasurementBuffer::~MeasurementBuffer(void)
 void MeasurementBuffer::BufferROSCloud(const sensor_msgs::PointCloud2& cloud)
 /*****************************************************************************/
 {
-  tryspatio_temporal_voxel_layer
+  try
   {
     pcl::PCLPointCloud2 pcl_pc2;
     pcl_conversions::toPCL(cloud, pcl_pc2);
--- a/src/measurement_buffer.cpp
+++ b/src/measurement_buffer.cpp
@@ -85,27 +85,6 @@ MeasurementBuffer::~MeasurementBuffer(void)
 void MeasurementBuffer::BufferROSCloud(const sensor_msgs::PointCloud2& cloud)
 /*****************************************************************************/
 {
-  try
-  {
-    pcl::PCLPointCloud2 pcl_pc2;
-    pcl_conversions::toPCL(cloud, pcl_pc2);
-    pcl::PointCloud < pcl::PointXYZ > pcl_cloud;
-    pcl::fromPCLPointCloud2(pcl_pc2, pcl_cloud);
-    BufferPCLCloud(pcl_cloud);
-  }
-  catch (pcl::PCLException& ex)
-  {
-    ROS_ERROR("Failed to convert to pcl type, dropping observation: %s", \
-              ex.what());
-    return;
-  }
-}
-
-/*****************************************************************************/
-void MeasurementBuffer::BufferPCLCloud(const \
-                                         pcl::PointCloud<pcl::PointXYZ>& cloud)
-/*****************************************************************************/
-{
   // add a new measurement to be populated
   _observation_list.push_front(observation::MeasurementReading());
 
@@ -123,7 +102,7 @@ void MeasurementBuffer::BufferPCLCloud(const \
     local_pose.pose.orientation.y=0;
     local_pose.pose.orientation.z=0;
     local_pose.pose.orientation.w=1;
-    local_pose.header.stamp = pcl_conversions::fromPCL(cloud.header).stamp;
+    local_pose.header.stamp = cloud.header.stamp;
     local_pose.header.frame_id = origin_frame;
 
     _buffer.canTransform(_global_frame, local_pose.header.frame_id , \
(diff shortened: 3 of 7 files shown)


include/spatio_temporal_voxel_layer/measurement_buffer.hpp
```diff
@@ -60,14 +62,13 @@
 #include <geometry_msgs/PoseStamped.h>
 // Mutex
 #include <boost/thread.hpp>
-#include <geometry_msgs/TransformStamped.h>
```
SteveMacenski: Removed the `TransformStamped` header?

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
```diff
@@ -153,7 +156,7 @@ class SpatioTemporalVoxelGrid
   int                             _decay_model;
   double                          _background_value, _voxel_size, _voxel_decay;
   bool                            _pub_voxels;
-  pcl::PointCloud<pcl::PointXYZ>::Ptr     _pc;
+  std::list<geometry_msgs::Point32>*   _grid_points;
```
SteveMacenski: I'm not sure how I feel about this. However if we're going to move to an STL container, it should probably be a `unordered_map` or `map` (unordered is slightly faster for iterating) and have the name as a `typedef` to abstract it

src/measurement_buffer.cpp
```diff
@@ -150,53 +129,46 @@ void MeasurementBuffer::BufferPCLCloud(const \
...
-    geometry_msgs::TransformStamped tf_stamped = _buffer.lookupTransform(_global_frame, cloud.header.frame_id.c_str(), pcl_conversions::fromPCL(cloud.header).stamp);
-    tf::StampedTransform tf_transform;
-    tf::transformMsgToTF(tf_stamped.transform, tf_transform);
-    tf_transform.frame_id_ = _global_frame;
-    tf_transform.stamp_ = tf_stamped.header.stamp;
-    pcl_ros::transformPointCloud(cloud, *cld_global, tf_transform);
-    cld_global->header.stamp = cloud.header.stamp;
+    // transform the cloud in the global frame
+    point_cloud_ptr cld_global(new sensor_msgs::PointCloud2());
+    geometry_msgs::TransformStamped tf_stamped = _buffer.lookupTransform(_global_frame, cloud.header.frame_id, cloud.header.stamp);
```
SteveMacenski: try to keep the line lengths inside the guards.

src/measurement_buffer.cpp
```diff
@@ -150,53 +129,46 @@ void MeasurementBuffer::BufferPCLCloud(const \
...
+    tf2::doTransform (cloud, *cld_global, tf_stamped);
```
SteveMacenski: Will `doTransform` also transfer the stamp? I know that was a concern before in line 160

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -85,6 +85,8 @@ void SpatioTemporalVoxelGrid::InitializeGrid(void)
   _grid->setName("SpatioTemporalVoxelLayer");
   _grid->insertMeta("Voxel Size", openvdb::FloatMetadata( _voxel_size ));
   _grid->setGridClass(openvdb::GRID_LEVEL_SET);
+
+  // setup PointCloud2
```
SteveMacenski: What's this here to indicate?

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -214,8 +216,11 @@ void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(const \
 
   if (_pub_voxels)
   {
-    _pc->push_back(pcl::PointXYZ(pose_world[0], pose_world[1], \
-                                 pose_world[2]));
+    geometry_msgs::Point32 point;
+    point.x=pose_world[0];
```
SteveMacenski: Add some spaces here just to be consistent

src/measurement_buffer.cpp
```diff
@@ -150,53 +129,46 @@ void MeasurementBuffer::BufferPCLCloud(const \
...
+        sor.setInputCloud (cloud_pcl);
+        sor.setFilterFieldName("z");
+        sor.setFilterLimits(_min_obstacle_height,_max_obstacle_height);
+        sor.setDownsampleAllData(false);
+        sor.setLeafSize ((float)_voxel_size,
+                         (float)_voxel_size,
+                         (float)_voxel_size);
+        sor.filter(*cloud_filtered);
+        cloud_filtered.swap(cloud_pcl);
+        cloud_pcl->is_dense=true;
```
SteveMacenski: What's this dense accomplish that didn't need to be here prior? I dont think you have a promised dense cloud on this output.

src/measurement_buffer.cpp
```diff
@@ -150,53 +129,46 @@ void MeasurementBuffer::BufferPCLCloud(const \
...
     }
 
-    // resize the cloud for the number of legal points
-    obs_cloud.points.resize(point_count);
-    obs_cloud.header.stamp = cloud.header.stamp;
-    obs_cloud.header.frame_id = cld_global->header.frame_id;
+    pcl_conversions::fromPCL(*cloud_pcl, *cld_global);
+    _observation_list.front()._cloud=cld_global;
```
SteveMacenski: Add spaces for `=` please

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -260,24 +265,30 @@ void SpatioTemporalVoxelGrid::operator()(const \
...
     const double cur_time = ros::WallTime::now().toSec();
 
-    pcl::PointCloud<pcl::PointXYZ>::const_iterator it;
-    for (it = obs._cloud->points.begin(); it < obs._cloud->points.end(); ++it)
+    const sensor_msgs::PointCloud2& cloud = *(obs._cloud);
+    sensor_msgs::PointCloud2ConstIterator<float> iter_x(cloud, "x");
+    sensor_msgs::PointCloud2ConstIterator<float> iter_y(cloud, "y");
+    sensor_msgs::PointCloud2ConstIterator<float> iter_z(cloud, "z");
+
+    for (; iter_x !=iter_x.end(); ++iter_x, ++iter_y, ++iter_z)
```
SteveMacenski: Please add `iter_x, iter_y, iter_z` just for completion

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -260,24 +265,30 @@ void SpatioTemporalVoxelGrid::operator()(const \
...
-      float distance_2 = (it->x - obs._origin.x) * (it->x - obs._origin.x) \
-                        + (it->y - obs._origin.y) * (it->y - obs._origin.y) \
-                        + (it->z - obs._origin.z) * (it->z - obs._origin.z);
-      if (distance_2 > mark_range_2 || distance_2 < 0.0001)
-      {
-        continue;
-      }
-      openvdb::Vec3d mark_grid(this->WorldToIndex( \
-                                       openvdb::Vec3d(it->x, it->y, it->z)));
+        float px = *iter_x, py = *iter_y, pz = *iter_z;
```
SteveMacenski: These can be references

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -260,24 +265,30 @@ void SpatioTemporalVoxelGrid::operator()(const \
...
+        float distance_2 = (px - obs._origin.x) * (px - obs._origin.x) \
+                          + (py - obs._origin.y) * (py - obs._origin.y) \
+                          + (pz - obs._origin.z) * (pz - obs._origin.z);
+        if (distance_2 > mark_range_2 || distance_2 < 0.0001)
+        {
+          continue;
+          std::cout << "Continuing." << std::endl;
```
SteveMacenski: Debug print?

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -260,24 +265,30 @@ void SpatioTemporalVoxelGrid::operator()(const \
...
+        }
+        openvdb::Vec3d mark_grid(this->WorldToIndex( \
+                                         openvdb::Vec3d(px, py, pz)));
 
-      if(!this->MarkGridPoint(openvdb::Coord(mark_grid[0], mark_grid[1], \
-                                             mark_grid[2]), cur_time))
-      {
-        std::cout << "Failed to mark point." << std::endl;
-      }
+        if(!this->MarkGridPoint(openvdb::Coord(mark_grid[0], mark_grid[1], \
```
SteveMacenski: It looks like this entire for loop was accidently indented in 4 rather than 2

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -330,12 +341,38 @@ double SpatioTemporalVoxelGrid::GetFrustumAcceleration( \
 }
 
 /*****************************************************************************/
-void SpatioTemporalVoxelGrid::GetOccupancyPointCloud( \
-                                       pcl::PointCloud<pcl::PointXYZ>::Ptr& pc)
+void SpatioTemporalVoxelGrid::GetOccupancyPointCloud(\
+                                       sensor_msgs::PointCloud2::Ptr& pc2)
```
SteveMacenski: Please align this with the guard  like before

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -330,12 +341,38 @@ double SpatioTemporalVoxelGrid::GetFrustumAcceleration( \
...
 /*****************************************************************************/
 {
-  // return the pointcloud stored
-  pc = _pc;
+  // convert the grid points stored in a PointCloud2
+  sensor_msgs::PointCloud2::Ptr grid_points_cloud (new sensor_msgs::PointCloud2());
```
SteveMacenski: You have this `typedef`ed, if you use it'll be less than max line length ;)

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -330,12 +341,38 @@ double SpatioTemporalVoxelGrid::GetFrustumAcceleration( \
...
+                                "x", 1, sensor_msgs::PointField::FLOAT32,
+                                "y", 1, sensor_msgs::PointField::FLOAT32,
+                                "z", 1, sensor_msgs::PointField::FLOAT32);
+  modifier.setPointCloud2FieldsByString(1, "xyz");
+
+  sensor_msgs::PointCloud2Iterator<float>iter_x(*grid_points_cloud, "x");
+  sensor_msgs::PointCloud2Iterator<float>iter_y(*grid_points_cloud, "y");
+  sensor_msgs::PointCloud2Iterator<float>iter_z(*grid_points_cloud, "z");
+
+  for(std::list<geometry_msgs::Point32>::iterator it = _grid_points->begin(); it != _grid_points->end(); ++it){
```
SteveMacenski: line length + typedef the list + brackets on a new line

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -330,12 +341,38 @@ double SpatioTemporalVoxelGrid::GetFrustumAcceleration( \
...
+    const geometry_msgs::Point32& pt = *it;
+    *iter_x = pt.x;
+    *iter_y = pt.y;
+    *iter_z = pt.z;
+    ++iter_x; ++iter_y; ++iter_z;
+  }
+
+  // return the pointcloud
+  pc2 = grid_points_cloud;
```
SteveMacenski: Why don't you use the already allocated `pc2` throughout the function?

src/measurement_buffer.cpp
```diff
@@ -150,53 +129,46 @@ void MeasurementBuffer::BufferPCLCloud(const \
...
-
-      // minimize information needed to process
-      cld_global->clear();
-      pcl::ApproximateVoxelGrid<pcl::PointXYZ> sor1;
-      sor1.setInputCloud (cld_no_nan);
-      sor1.setLeafSize ((float)_voxel_size,
-                        (float)_voxel_size,
-                        (float)_voxel_size);
-      sor1.filter (*cld_global);
+        pcl::VoxelGrid<pcl::PCLPointCloud2> sor;
```
SteveMacenski: 4 spaces?

src/measurement_buffer.cpp
```diff
@@ -150,53 +129,46 @@ void MeasurementBuffer::BufferPCLCloud(const \
...
+        sor.setInputCloud (cloud_pcl);
+        sor.setFilterFieldName("z");
```
SteveMacenski: This height filter is probably the reason you get a performance increase, since this is a relatively heavy operation.

src/measurement_buffer.cpp
```diff
@@ -150,53 +129,46 @@ void MeasurementBuffer::BufferPCLCloud(const \
...
-      {
-        obs_cloud.points.at(point_count++) = *it;
-      }
+        pcl::PassThrough<pcl::PCLPointCloud2> pass_through_filter;
+        pass_through_filter.setInputCloud(cloud_pcl);
+        pass_through_filter.setKeepOrganized(false);
+        pass_through_filter.setFilterFieldName("z");
+        pass_through_filter.setFilterLimits(_min_obstacle_height,_max_obstacle_height);
+        pass_through_filter.filter(*cloud_filtered);
+        cloud_filtered.swap(cloud_pcl);
```
SteveMacenski: Is there a reason you swap pointers vs. just operating on the same pcl pointer? I dont see much benefit from having 2 pointers and one is just being copied out

include/spatio_temporal_voxel_layer/measurement_buffer.hpp
doisyg: Putting it back (i don't know how but it was compiling without)

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
doisyg: I used the same approach as in the PR you pointed: https://github.com/ros-planning/navigation/pull/755/files
I did not change to unordered_map yet, but I have no problem with switching (side note, it is only used for visualization)

src/measurement_buffer.cpp
doisyg: Ok, changed it manually. Do you use any auto formatting tools that you would recommend to keep consistency with your code?

src/measurement_buffer.cpp
doisyg: Yes, this is much cleaner as we can natively transform the pointcloud with TF2 (since it is in the ROS sensor_msgs::PointCloud2 format now)

src/spatio_temporal_voxel_grid.cpp
doisyg: Removed

src/spatio_temporal_voxel_grid.cpp
doisyg: Added

src/measurement_buffer.cpp
doisyg: Output of the filter should be free of NaN, hence dense. I set the flag to be sure it is treated as a dense cloud after (which may accelerated the conversion, not sure what's going on under the hood).
But I just checked, the flag is already set to true after the filter, so removing.

src/measurement_buffer.cpp
doisyg: Added

src/spatio_temporal_voxel_grid.cpp
doisyg: Added

src/spatio_temporal_voxel_grid.cpp
doisyg: Actually `*iter_x, *iter_y, *iter_z`, can be used directly after, so I removed the px, py, pz declaration

src/spatio_temporal_voxel_grid.cpp
doisyg: Removed

src/spatio_temporal_voxel_grid.cpp
doisyg: Corrected

src/spatio_temporal_voxel_grid.cpp
doisyg: Ok

src/spatio_temporal_voxel_grid.cpp
doisyg: Where? I corrected the line length anyway

src/spatio_temporal_voxel_grid.cpp
doisyg: Corrected line length + bracket. Not a huge fan of using too much typedef as I can get confused about the the type later. But if you insist, I can do it.

src/spatio_temporal_voxel_grid.cpp
doisyg: Good point

src/measurement_buffer.cpp
doisyg: Bad configuration of my IDE, fixed

src/measurement_buffer.cpp
doisyg: + having to deal with simpler conversions

src/measurement_buffer.cpp
doisyg: PCL filters requires to have an input and an output cloud pointer.
However I now removed the swap and used cloud_filtered as an input of the next conversion (reason for the swap was having the possibility to comment the whole filter block during testing).

src/measurement_buffer.cpp
SteveMacenski: I don't -- frankly I'm not too picky as long as its consistent and that's the style I had been working with prior. Generally my philosophy is that it should stay within a 100 line guard when reasonable, and any subsequent line extensions should be aligned with that end guard as well.

src/measurement_buffer.cpp
SteveMacenski: Ah thanks for the explanation, I assumed dense has something to do with spatial filled-ness, not that its all finite. Learn something new every day.

src/spatio_temporal_voxel_grid.cpp
SteveMacenski: I think typedefs make alot of sense if you consistently reuse a templated class over and over but still only use 1 unique template type. I'd agree if we called std::vector<int> vs double, float, Grid, etc but given that the pointcloud is a list of a set type, it makes it clearer to read. Also that way if we want to swap out `list` for `vector` or `map` etc, its a 1 line change ;)

SteveMacenski:
Thanks, went through my comments and seems like all were addressed, I'll do another run through later today. In the meantime getting to root cause of your PCL issues would be good so we know what the issue is that caused the need for this. If at all possible I prefer to use PCL for the flexibility of future features but I'm not sure the scope of this particular project will warrant it. 

On a quick pass I see some (new?) debug artifacts
```
  ros::Time t1, t2;
  ros::Duration t_diff;
```
that are unused, and tab vs spaces issues in the filters section and the `distance_2` section (if you look at the Github view - spaces are preferred).
I'd recommend doing a once-over on the GitHub view and seeing some of the formatting issues

doisyg:
Just removed the debug artifacts.
There is no rush on my side to see this merged and released. I am fine with this PR left open for tests and discussions by others for some time.
I believe that independently from the performance issue I experienced, it is more consistent to avoid transformation to the apparently old `pcl::PointCloud<pcl::PointXYZ>` format (or at least as late as possible in the proceeding pipeline), in a similar ways as what was done in the core ROS pacakges: for instance https://github.com/ros-planning/navigation/blob/melodic-devel/base_local_planner/include/base_local_planner/voxel_grid_model.h
This PR is not getting rid of PCL, but base all pointcloud operation on the `sensor_msgs/PointCloud2` / `pcl::PCLPointCloud2` formats. Most PCL functions are compatible with the `pcl::PCLPointCloud2` format.

SteveMacenski:
Well I know it was the maintainers view that getting rid of PCL from the navigation stack was something important to do. I don't personally fully understand the rationale but certainly less dependencies is good. I was not aware that `pcl::PointCloud<PointT>` is old, I thought the `PCLPointCloud2` was only a convenience tool for embedded ROS users.

This is true. Even if root cause isn't found its a good contribution in its own right if that's the direction the community wants to go, I certainly won't be the one to stand in the way :-) 

I'm going to try to play some games with some robots I have at home and see if I get one's stack online in Melodic once my new shiny Jetson Nano comes in the mail (which only runs Melodic). At that point I can be far more constructive here in things I'm seeing in Melodic

SteveMacenski:
In the meantime, i want to get this formatting stuff squared away so once we have a couple of independent verification we can ship it.

src/spatio_temporal_voxel_grid.cpp
doisyg: Not relevant anymore since we use the input pointer

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
doisyg: Coming back on this point, I don't understand why you want to use `map` or `unordered_map`  to store `geometry_msgs::Point32` objects. Why using need associative containers here?
However, one can argue `vector` could be faster than `list` for this use case. https://baptiste-wicht.com/posts/2012/12/cpp-benchmark-vector-list-deque.html

SteveMacenski:
@doisyg have you been running this for the last N weeks since we last discussed? If this is stable, I'm fairly open to merging it now.

doisyg:
I know @Patrick-Lascombe used it extensively for the past few weeks, maybe he can comment (and also once its merged, he can PR some of the improvement he's been working on).
Patrick?

SteveMacenski:
Sweet, I look forward to it. I've also been thinking about how to make the spatial component stronger. The clearing I think is a little too aggressive for robots without alot of sensor coverage. I've thought a bit about adding a heuristic to the clearing so that things within an envelope of the robot have greater persistence.

doisyg:
Looking forward to it too! I have been quite busy with other projects but I will try to dive in again soon.

nickvaras:
In with the last point, I never got to implement the "no-decay" volumes I
originally wanted, and although the standard approach works ok in most
cases, there's been been situations where total persistence of obstacles
outside FOV has been needed... so expect PR hopefully sooner rather than
later.

Patrick-Lascombe:
The layer has been working really well for the past weeks. This PR can be merged now and as doisyg said I'll be PRing improvements when they are stable on my side.

SteveMacenski:
Ok @Patrick-Lascombe @doisyg I'm going to do another review to get the rest of the stuff resolved and then I'll merge. Its mostly formatting, I think someone's code editor might have some pythonic settings ;)

src/measurement_buffer.cpp
```diff
@@ -150,53 +129,44 @@ void MeasurementBuffer::BufferPCLCloud(const \
...
-    geometry_msgs::TransformStamped tf_stamped = _buffer.lookupTransform(_global_frame, cloud.header.frame_id.c_str(), pcl_conversions::fromPCL(cloud.header).stamp);
-    tf::StampedTransform tf_transform;
-    tf::transformMsgToTF(tf_stamped.transform, tf_transform);
-    tf_transform.frame_id_ = _global_frame;
-    tf_transform.stamp_ = tf_stamped.header.stamp;
-    pcl_ros::transformPointCloud(cloud, *cld_global, tf_transform);
-    cld_global->header.stamp = cloud.header.stamp;
+    // transform the cloud in the global frame
+    point_cloud_ptr cld_global(new sensor_msgs::PointCloud2());
+		geometry_msgs::TransformStamped tf_stamped = _buffer.lookupTransform( \
```
SteveMacenski: This  section is weird, please remove all indentations and replace with spaces.

src/measurement_buffer.cpp
```diff
@@ -150,53 +129,44 @@ void MeasurementBuffer::BufferPCLCloud(const \
...
-
-      // minimize information needed to process
-      cld_global->clear();
-      pcl::ApproximateVoxelGrid<pcl::PointXYZ> sor1;
-      sor1.setInputCloud (cld_no_nan);
-      sor1.setLeafSize ((float)_voxel_size,
-                        (float)_voxel_size,
-                        (float)_voxel_size);
-      sor1.filter (*cld_global);
+			pcl::VoxelGrid<pcl::PCLPointCloud2> sor;
```
SteveMacenski: Here as well, these should be spaces (and only 2)

src/measurement_buffer.cpp
```diff
@@ -150,53 +129,44 @@ void MeasurementBuffer::BufferPCLCloud(const \
...
-    unsigned int point_count = 0;
-    pcl::PointCloud<pcl::PointXYZ>::iterator it;
-    for (it = cld_global->begin(); it != cld_global->end(); ++it)
+    else
     {
-      if (it->z <= _max_obstacle_height && it->z >= _min_obstacle_height)
-      {
-        obs_cloud.points.at(point_count++) = *it;
-      }
+			pcl::PassThrough<pcl::PCLPointCloud2> pass_through_filter;
```
SteveMacenski: tabs

src/measurement_buffer.cpp
```diff
@@ -252,7 +222,7 @@ void MeasurementBuffer::RemoveStaleObservations(void)
   {
     observation::MeasurementReading& obs = *it;
     const ros::Duration time_diff = \
-            _last_updated - pcl_conversions::fromPCL(obs._cloud->header).stamp;
+            _last_updated - obs._cloud->header.stamp;
```
SteveMacenski: This can be on 1 line now

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -214,8 +215,11 @@ void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(const \
 
   if (_pub_voxels)
   {
-    _pc->push_back(pcl::PointXYZ(pose_world[0], pose_world[1], \
-                                 pose_world[2]));
+    geometry_msgs::Point32 point;
+		point.x = pose_world[0];
```
SteveMacenski: tabs

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -260,24 +264,30 @@ void SpatioTemporalVoxelGrid::operator()(const \
...
+    sensor_msgs::PointCloud2ConstIterator<float> iter_x(cloud, "x");
+    sensor_msgs::PointCloud2ConstIterator<float> iter_y(cloud, "y");
+    sensor_msgs::PointCloud2ConstIterator<float> iter_z(cloud, "z");
 
-      if(!this->MarkGridPoint(openvdb::Coord(mark_grid[0], mark_grid[1], \
-                                             mark_grid[2]), cur_time))
-      {
-        std::cout << "Failed to mark point." << std::endl;
-      }
+		for (iter_x, iter_y, iter_z; iter_x !=iter_x.end(); \
```
SteveMacenski: tabs

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -331,11 +341,36 @@ double SpatioTemporalVoxelGrid::GetFrustumAcceleration( \
 
 /*****************************************************************************/
 void SpatioTemporalVoxelGrid::GetOccupancyPointCloud( \
-                                       pcl::PointCloud<pcl::PointXYZ>::Ptr& pc)
+                                            sensor_msgs::PointCloud2::Ptr& pc2)
 /*****************************************************************************/
 {
-  // return the pointcloud stored
-  pc = _pc;
+  // convert the grid points stored in a PointCloud2
```
SteveMacenski: this function is very strange now, please realign with 2 spaces a scope

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -166,6 +166,7 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
     // clears measurement buffer after reading values from it
     source_node.param("clear_after_reading", clear_after_reading, false);
 
+
```
SteveMacenski: remove this line please

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -328,10 +329,10 @@ void SpatioTemporalVoxelLayer::PointCloud2Callback( \
                 const boost::shared_ptr<buffer::MeasurementBuffer>& buffer)
 /*****************************************************************************/
 {
-  // buffer the point cloud
-  buffer->Lock();
-  buffer->BufferROSCloud(*message);
-  buffer->Unlock();
+	// buffer the point cloud
```
SteveMacenski: revert back to 2 spaces

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
SteveMacenski: Lets change this to an `unordered_set` that gives us faster random access.

doisyg:
> Ok @Patrick-Lascombe @doisyg I'm going to do another review to get the rest of the stuff resolved and then I'll merge. Its mostly formatting, I think someone's code editor might have some pythonic settings ;)

Indeed, I had tabs being displayed at two spaces (but still inserted as tabs) at the time of the last edit. Pushed the formating changes

doisyg:
I did not change `std::list<geometry_msgs::Point32>* _grid_points` to an `unordered_set`. Can you make a case for it? To my understanding we don't need any random access to `_grid_points` (and `std::list` doesn't allow random access anyway)

SteveMacenski:
Sure, a `list` is a doubly-linked list:
```
std::list is a container that supports constant time insertion and removal of elements from anywhere in the container. Fast random access is not supported. It is usually implemented as a doubly-linked list.
```

Whereas unordered set is a hash table
```
Unordered set is an associative container that contains a set of unique objects of type Key. Search, insertion, and removal have average constant-time complexity. 
```

Such that we get O(1) (assuming no collisions) non-sequential access. Right now we're not actually doing that -- but its preventative. With that said, sure, we can just not do it here, I can do later if it really bothers me. 

In doing research on "when to use std::list" I've come to the conclusion that that isn't really the right data structure, `unordered_set` could work but also probably isnt either since we're iterating through it sequentially. It really should be a `vector`

```
[use lists] Situations where you want to insert a lot of items into anywhere but the end of a sequence repeatedly.
```

We're adding them all at the end, so that O(1) insertion anywhere isn't really relevant to us. 

Thoughts?

doisyg:
Thanks for your detailed opinion

> Such that we get O(1) (assuming no collisions) non-sequential access. Right now we're not actually doing that -- but its preventative. With that said, sure, we can just not do it here, I can do later if it really bothers me.

Agreed, definitely worth to drop the list if there is a chance that we need a different use case for this object in the future. There is the risk though that with vector we pay an additional cost at the insertion when there is no available contiguous memory, but frankly this is subject to debate and it only happens occasionaly (and we can always preallocate if needed).

So switching to vector ;) I just pushed the change. I have no robot with me et the moment so I could not test more than compilation.

SteveMacenski:
Ok, when you can do a quick test saying "cool still works, no real CPU change" go ahead and squash merge!

Patrick-Lascombe:
Sorry for the late answer, I tested it and it still works with good performances ! 
You can merge it

SteveMacenski:
@Patrick-Lascombe @doisyg awesome job, I'm cherry picking all the good stuff from Kinetic over to Melodic now. I'll submit another PR for that - I'd love if you were able to just give me the thumbs up that it stills still at a high level so I can release on bloom for you guys.

You'll also get some shiny new features like: 3D lidar support, dynamic reconfigure, and multiple stream declarations in Melodic afterwards as well

doisyg:
Awesome! We will test it asap

PR #122, opened 2019-03-11
created clear area method to be able to properly clear  spatio layer from the costmap

MarcelSoler:
Still to be tested.

Pending things:
 - Implement a way to clear not the whole grid but the selected zone. I am not familiar with openvdb I have written some pseudocode (l.647).

 But i have two main doubts : 
- Probably a for loop to select all z points of the grid is not necessary and there is faster an cheaper method to select the whole column of x,y points.
-  Is my convertion (l.661) from costmap to grid coordinates correct?

### DIFF
--- a/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
+++ b/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
@@ -96,6 +96,8 @@ public:
   virtual void updateCosts(costmap_2d::Costmap2D& master_grid, int min_i, int min_j, \
                                                                int max_i, int max_j);
 
+  virtual void clearArea(int start_x, int start_y, int end_x, int end_y) override;
+
   // Functions to interact with other layers
   virtual void matchSize(void);
 
--- a/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
+++ b/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
@@ -96,7 +96,7 @@ public:
   virtual void updateCosts(costmap_2d::Costmap2D& master_grid, int min_i, int min_j, \
                                                                int max_i, int max_j);
 
-  virtual void clearArea(int start_x, int start_y, int end_x, int end_y) override;
+  virtual void clearArea(int start_x, int start_y, int end_x, int end_y);
 
   // Functions to interact with other layers
   virtual void matchSize(void);
--- a/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
+++ b/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
@@ -43,9 +43,11 @@
 
 // voxel grid
 #include <spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp>
+#include <spatio_temporal_voxel_layer/SpatioTemporalVoxelLayerConfig.h>
 // ROS
 #include <ros/ros.h>
 #include <message_filters/subscriber.h>
+#include <dynamic_reconfigure/server.h>
 // costmap
 #include <costmap_2d/layer.h>
 #include <costmap_2d/layered_costmap.h>
@@ -64,6 +66,7 @@
 #include <sensor_msgs/point_cloud_conversion.h>
 #include <geometry_msgs/Point.h>
 #include <spatio_temporal_voxel_layer/SaveGrid.h>
+#include <std_srvs/SetBool.h>
 // projector
 #include <laser_geometry/laser_geometry.h>
 // tf
@@ -76,6 +79,8 @@ namespace spatio_temporal_voxel_layer
 // conveniences for line lengths
 typedef std::vector<boost::shared_ptr<message_filters::SubscriberBase> >::iterator observation_subscribers_iter;
 typedef std::vector<boost::shared_ptr<buffer::MeasurementBuffer> >::iterator observation_buffers_iter;
+typedef spatio_temporal_voxel_layer::SpatioTemporalVoxelLayerConfig dynamicReconfigureType;
+typedef dynamic_reconfigure::Server<dynamicReconfigureType> dynamicReconfigureServerType;
 
 // Core ROS voxel layer class
 class SpatioTemporalVoxelLayer : public costmap_2d::CostmapLayer
@@ -98,6 +103,7 @@ public:
   virtual void reset(void);
   virtual void activate(void);
   virtual void deactivate(void);
+
 
   // Functions for sensor feeds
   bool GetMarkingObservations(std::vector<observation::MeasurementReading>& marking_observations) const;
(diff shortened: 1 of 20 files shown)


SteveMacenski:
- Shouldn't there be a service or similar to access that method to clear?
- If you look at how I populate the pointcloud for visualization, you can see how to efficiently access the voxel grid. Since the grid is *sparse* the method probably wouldn't work and would be very time consuming.  
- Personally, I don't prescribe to the "clearing areas" methodology. If you think your environmental model is incorrect or needing clearing, go nuclear and clear it all. I've never found the default recovery behavior with a distance to be better than clearing the full costmap if necessary. I think this is more true here since you're not storing data about places 30 minutes since passed to be concerned about clearing out. Can you describe why you think this is important, maybe I'm missing something. I viewed this as one of those legacy Willow items that I was happy to do away with.

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -636,6 +636,44 @@ void SpatioTemporalVoxelLayer::updateCosts( \
...
+/*****************************************************************************/
+void SpatioTemporalVoxelLayer::clearArea(int start_x, int start_y, int end_x, int end_y)
+/*****************************************************************************/
+{
+  //easy way reset the whole grid.
+  _voxel_grid->ResetGrid();
+  Costmap2D::resetMaps();
+
+
+  // hard way selected points
```
SteveMacenski: In general, I dont like to commit commented out code into a master branch

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
```diff
@@ -96,6 +96,8 @@ class SpatioTemporalVoxelLayer : public costmap_2d::CostmapLayer
   virtual void updateCosts(costmap_2d::Costmap2D& master_grid, int min_i, int min_j, \
                                                                int max_i, int max_j);
 
+  virtual void clearArea(int start_x, int start_y, int end_x, int end_y) override;
```
SteveMacenski: What is this overriding?

SteveMacenski (changes requested):

MarcelSoler:
> * Shouldn't there be a service or similar to access that method to clear?

I still have to test, but I think that clear costmap service will clear also the spatio layer. I can still create a service to do so. I'll test it in Thursday.

> * If you look at how I populate the pointcloud for visualization, you can see how to efficiently access the voxel grid. Since the grid is _sparse_ the method probably wouldn't work and would be very time consuming.
Ok.

> * Personally, I don't prescribe to the "clearing areas" methodology. If you think your environmental model is incorrect or needing clearing, go nuclear and clear it all. I've never found the default recovery behavior with a distance to be better than clearing the full costmap if necessary. I think this is more true here since you're not storing data about places 30 minutes since passed to be concerned about clearing out. Can you describe why you think this is important, maybe I'm missing something. I viewed this as one of those legacy Willow items that I was happy to do away with.

I completely agree. I rather clear the full costmap.

src/spatio_temporal_voxel_layer.cpp
MarcelSoler: Sure I will remove it. It was an easy way of sharing my thoughts.

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
MarcelSoler: Are you referring at the override identifier? 
It a personal preference,  it can be removed. It tells explicitly to the compiler that is overriding a virtual function of the base class. This way it will fail in compilation is the base method doesn't have the same signature or if it doesn't exist.

SteveMacenski:
If you're OK not clearing out areas of the costmap, then the pseudo-code can definitely be removed. I always like to clear the full grid if there's a reason to clear it at all. Half measures lead to half result. 

The clear costmap service already _did_ clear this costmap. The only thing that doesnt is the clear costmap recovery behaviors. Are you referencing my fork (https://github.com/ros-planning/navigation/compare/kinetic-devel...SteveMacenski:layer_clear_area) with this PR having the `clearArea` method? If so, that's not been merged yet and you also still need to make sure that you're clearing out the costmap grid, so you should still call the base implementation of `clearArea` as well so that you're clearing the costmap and the internal structure

This is why I asked about the override/virtual method, since there's currently nothing to override :-). Fergs will probably merge it for Melodic, but he's being hesitant on the kinetic branch since its more in "maintenance mode"

MarcelSoler:
> If you're OK not clearing out areas of the costmap, then the pseudo-code can definitely be removed. I always like to clear the full grid if there's a reason to clear it at all. Half measures lead to half result.

Done!

> The clear costmap service already _did_ clear this costmap. The only thing that doesnt is the clear costmap recovery behaviors. Are you referencing my fork ([ros-planning/navigation@kinetic-devel...SteveMacenski:layer_clear_area](https://github.com/ros-planning/navigation/compare/kinetic-devel...SteveMacenski:layer_clear_area)) with this PR having the `clearArea` method? If so, that's not been merged yet and you also still need to make sure that you're clearing out the costmap grid, so you should still call the base implementation of `clearArea` as well so that you're clearing the costmap and the internal structure

Yes I am referencing your fork. I am not sure if I still have to use the base implementation of clearArea.
With the recovery behavour I am clearing two layers:
- Lidars one.
- Spatio temporal one.

Here attached my recovery configuration:
```
recovery_behaviours:  [{name: aggressive_reset, type: clear_costmap_recovery/ClearCostmapRecovery}]

aggressive_reset:
  reset_distance: 0.0
  layer_names: ["obstacles", "spatio_temporal"]
```
For what I understand, obstacles layer will be clear by the base implementation.
And the spatio temporal will be cleared with clear area override, right?

```
void SpatioTemporalVoxelLayer::clearArea(int start_x, int start_y, int end_x, int end_y)
{
  _voxel_grid->ResetGrid();
  Costmap2D::resetMaps();
}
```
 `Costmap2D::resetMaps();` will reset the contribution of spatio_temporal_layer on the costmap right? If not I will need call the base implementation of `clearArea` as well.


Ok for now I'll test it, with your kinetic branch and let's see how it evolves. Eventually I'll pass to melodic but i don't know the exact timelines...

SteveMacenski:
I think you may have not pushed your code to GitHub, I haven't seen the update. 

Base implementation would deal with everything not-STVL. STVL you'd need to clear out the data structure (done) but also still do the same thing as the base implementation to clear out section of the costmap (if that's wanting to be supported), else maybe resetMaps does that as well, I'm not 100% certain of that fact.

MarcelSoler:
Push failed no internet connection... Sorry for the confusion I was really sure I had done it.

I'll test this commits, try reset maps to see if it will works as I expected it to. 

Form my side I can wait to see how the navigation pull request evolves, also I can create the custom recovery.
In my opinion from the user point of view the behaviour is clearer without a custom recovery. However for the sake of compatibility and version control  between kinetic melodic maybe it's recommendable that both versions work with the same implementation.

SteveMacenski:
@MarcelSoler any updates? I'm willing to merge this as Part 1 of 2 part PRs if there's a recovery that will use this in part 2

MarcelSoler:
no updates :( I am hoping to test it this week.

MarcelSoler:
I couldn't test in kinetic... I am hoping to test it in melodic by the end of this week or beginning of next one...

SteveMacenski:
No problem, just retarget the PR branch

SteveMacenski:
[old old old....] hey there

What are you using these days? If you're on melodic, we can just merge in melodic, Kinetic shouldnt be messed around with too much anyhow. You have some merge conflicts to deal with to make that happen (might be worth rebasing / cherry picking)

We can merge this in if you can add the recovery -- I'm going to push harder on the navigation1 guys to merge this into melodic

MarcelSoler:
I am using melodic stable branch.

The problem I encountered: my robot cleans the cosmap if path obstructed. The problem appears when the path is obstructed because some 3D obstacle, my robot has blind spots. Hence if I clear the whole spatio temporal layer I could crash into something...

I still want to finish this pull request, however since is not critical development right now, I am having problems allocating resources to do it... I wanted to gain time, waiting navigation pull request to be merged in Melodic.

SteveMacenski:
Ok. I will also push on @dlu to help get that over the line. That update is cleaner anyway. 

I have control over navigation2 so that’ll be easy to support come ros2 time.

SteveMacenski:
Wow, that was fast, its in navigation now.

MarcelSoler:
Ancient PR I am not using navigation stack any more 😢 . So I cannot follow up on this...
Should we close it? Sorry for the trouble...

SteveMacenski:
Perhaps so given the massive list of merge conflicts now! Otherwise, I would have just kept it in case it helped someone in the future

PR #125, opened 2019-05-31
Fix melodic compilation

### LINKED ISSUE #124: Melodic-devel error on lambda function
Following the #123, I had the following errors that appears in the DynamicReconfigureCallback
```
/home/patrick/stvl_ws/src/spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp:586:45: error: use of ‘auto’ in lambda parameter declaration only available with -std=c++14 or -std=gnu++14
   auto updateFlagIfChanged = [&update_grid](auto& own, const auto& reference)
                                             

/home/patrick/stvl_ws/src/spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp:586:62: error: use of ‘auto’ in lambda parameter declaration only available with -std=c++14 or -std=gnu++14
   auto updateFlagIfChanged = [&update_grid](auto& own, const auto& reference)
```
My g++ version is 7.3.0 so it should have c++14/c++17.
I do not have a lot of experience concerning lamda functions, could it be a requirement or something that my computer is missing ?

Thanks for your help

MarcelSoler:
Motivation:
Fix #124 , getting rid of lamdas.

Explanation:
Lamdas are not really necessary in this function, dynamic reconfigure has a way to know what group of parameter has been modified, a bitmask called level.
The inspiration idea credit comes from my colleagues from  [beta_robots](http://www.beta-robots.com/) .

Changes made:
- Separated reconfigure parameters in 2 levels [0,1].
- If any parameter level 1 is modified, in the
callback the received level value will be 1 and we'll update grid.


MarcelSoler:
@SteveMacenski  tested ok from my side!

cfg/SpatioTemporalVoxelLayer.cfg
```diff
@@ -15,15 +15,15 @@ decay_model.append(gen.const("Permanent", int_t, 2, "Permanent"))
 decay_model_enum = gen.enum(decay_model, "Decay models")
 
 gen.add("enabled", bool_t, 0, "Enabled or not", True)
-gen.add("publish_voxel_map", bool_t, 0, "Publishes the voxel map", False)
-gen.add("voxel_size", double_t, 0, "Size of the voxel in meters", 0.05, 0.001, 10.0)
+gen.add("publish_voxel_map", bool_t, 1, "Publishes the voxel map", False)
```
SteveMacenski: Why did you change the defaults? I would not like to have the voxel map auto publish -- and making the others `0` makes it clearer to me when those are overwriting real parameters or not (because size 0 is nonsensical and would crash, helps debug)

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -582,35 +582,24 @@ void SpatioTemporalVoxelLayer::DynamicReconfigureCallback( \
...
   _enabled = config.enabled;
   _combination_method = config.combination_method;
   _mark_threshold = config.mark_threshold;
   _update_footprint_enabled = config.update_footprint_enabled;
   _mapping_mode = config.mapping_mode;
   _map_save_duration = ros::Duration(config.map_save_duration);
-  _decay_model = static_cast<volume_grid::GlobalDecayModel>(decay_model_int);
 
-  if (update_grid)
+  if (level >=1) //update grid
```
SteveMacenski: So, I've never used the level parameter before, so please educate me on why this would tell you if you need to update just those parameters and make a new voxel grid? The rationale behind that function was so that if things changed that requires re-instantiating the underlying grid we create a new one only when those parameters change. I think the level indicator wouldn't be the discriminator here.

SteveMacenski: Ah, I think I understand, the level will change if 0->1 or 1->0's from the defaults. 

If that's true, what happens when you change something you changed once back down to the original? I dont think that would re-instantiate the grid object because the level would look like its opposite case - for instance as you would while parameter tuning.

cfg/SpatioTemporalVoxelLayer.cfg
MarcelSoler: The third parameter is the level not the default value. https://wiki.ros.org/dynamic_reconfigure/Tutorials/HowToWriteYourFirstCfgFile

I have set up two levels 0 , 1.  1 for the parameters affecting the grid.

src/spatio_temporal_voxel_layer.cpp
MarcelSoler: For what I understand each time reconfigure is called level parameter changes according which group of params affects. 
- If only fields with level  0 change level will be 0.
- If only fields with 1 value change level will be 1.
- If both levels field are updated level will be 0 + 1 = 1 .

Level works as a bitmask to keep track of which group of parameters has changed.
If when reconfigure callback is called level is 0 we know that all grid parameters haven't changed, so we don't update the grid. If level is 1 we update the grid.

SteveMacenski: Got it, my bad, I thought this was defaults from my quick scanning. This makes sense

SteveMacenski:
Great work, as always. I appreciate the help!

PR #127, opened 2019-06-06
fix undefined symbol to tf2::getTimestamp(geometry_msgs::PoseStamped)

dorezyuk:
Hi,
I had undefined symbol issues to the tf2::getTimestamp(geometry_mgs::PoseStamped). 

This is intrinsically used by the tf2_ros::Buffer::transform, when the function is called with geometry_msgs::PoseStamped as argument. The callee is inside measurement_buffer.cpp

Technically the callee is implemented correctly, and it is rather a design flaw in the tf2_ros::Buffer; However this is a pragmatic fix.

I also included the tf_geometry_msgs dependency, used in the spatio_temporal_voxel_layer.cpp. This package is not inside of the dependencies of tf2_ros, hence must be included explicitly.

Best, 
Dima

// edit, typos

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,7 @@ find_package(catkin REQUIRED COMPONENTS
   pcl_ros
   roscpp
   tf2_ros
+  tf2_geometry_msgs
   visualization_msgs
   message_generation
   dynamic_reconfigure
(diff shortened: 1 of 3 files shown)


SteveMacenski:
Awesome! Thanks for the additions.

PR #128, opened 2019-06-11
Replace double-based hashing with index-based hashing

dorezyuk:
# Idea

currently the spvl tries to hash floating point values in order to store them in a hash map. Since hashing floating point values is bad practise, this commit implements hashing based on a continuous index. 

# Details

converter_func is a wrapper abstracting how world coordinates will be converted into a 1D index.

# Testing

Code run successfully on one of magazino's custom robots

# Performace

Modified code showed additionally a slight decrease in cpu/mem footprint

### DIFF
--- a/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
+++ b/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
@@ -45,6 +45,7 @@
 #include <pcl/PCLPointCloud2.h>
 // ROS
 #include <ros/ros.h>
+#include <costmap_2d/costmap_2d.h>
 // STL
 #include <math.h>
 #include <unordered_map>
@@ -81,22 +82,6 @@ enum GlobalDecayModel
   PERSISTENT = 2
 };
 
-// Structure for an occupied cell for map
-struct occupany_cell
-{
-  occupany_cell(const double& _x, const double& _y) :
-    x(_x), y(_y)
-  {
-  }
-
-  bool operator==(const occupany_cell& other) const
-  {
-    return x==other.x && y==other.y;
-  }
-
-  double x, y;
-};
-
 // Structure for wrapping frustum model and necessary metadata
 struct frustum_model
 {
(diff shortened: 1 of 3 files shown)


SteveMacenski:
I’m curious: how “slight” is slight? 

Currently on vacation- will address this when I’m back state-side

dorezyuk:
On the system I’ve tested it cpu went down from 33 to 28 and mem from 1.9 to 1.8. But this tells very little so it would best you cross check that on your setup

dorezyuk:
If I have time today I will try to run both versions longer, since I somehow believe that the first measurement is more due to noise

SteveMacenski:
Ok, i’d like some same-data comparisons to support. You make some _significant_ changes to design decisions and I want to make sure the trade offs make sense. Including ROS specifics in the grid itself was a very intentional separation of capabilities vs. wrappers.

dorezyuk:
Hi Steve,  
thanks for your reply. Here are some points

1. Runtime
please check out the sample code attached; Don't mind the odd naming (its just due to github)
[hash_map.cpp.txt](https://github.com/SteveMacenski/spatio_temporal_voxel_layer/files/3303180/hash_map.cpp.txt) It contains samples of both approaches. For running it you will need the benchmark lib from google. The test emulates only the write process to the hash map; The read process will be in both cases mostly identical.
The results are:

> Benchmark           Time             CPU            Iterations 
> BM_steve        35203 ns        35203 ns        19295
> BM_dima         16456 ns        16456 ns        41682

2. Structure
I do like your effort to keep the openvdb and ROS stuff separated. And its true, my commit is mixing it. However, the entire code is a plugin for a ROS library. So keeping ROS out of some parts may or may not be reasonable. But anyhow, if you don't like the mixing, we could either rebuild the world-to-index conversion or use another conversion

3. Safety
As mentioned above, hashing floating points may return undesired results, depending on platforms, compilers, etc (see [example1](https://stackoverflow.com/questions/4238122/hash-function-for-floats), but you can just search yourself for "c++ hash float"). You are checking how many voxels fall into the same [grid-cell](https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/1ce9845216dac4065ffe83424dc7ba1bf2ca6cfb/src/spatio_temporal_voxel_layer.cpp#L670) before marking. A flawed floating point hash might lead to occupied grid cell not being marked correctly, making your code not suited for those users.

Best ,
Dima

SteveMacenski:
An easier way to get rid of double hashing is to use the OpenVDB internal node positions rather than the 3D xyz positions. That’ll not mess with the structure and be based on integer positions. 

The internal positions are analogous to the costmap 2d worldToGrid but not necessarily the same depending on setup since this is actually represented as an infinite grid.

PR #133, opened 2019-07-18
Fix the build on debian stretch.

clalancette:
There is a transitive dependency on Qt5::Widgets, which causes
the build to fail.  Remove the transitive dependencies, which
allows it to compile.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>


SteveMacenski:
Your great, thanks for the help!

PR #142, opened 2019-10-21
Exposes enabled parameter for frustums

nickovaras:
This is to allow some frustums not to be enabled at startup, and to be only selectively enabled or disabled at runtime.


SteveMacenski:
that was fast, please also submit this to melodic to keep in sync

nickovaras:
Thanks!!!

SteveMacenski:
@nickovaras please also submit a PR updating the example configuration files with the enable parameter in it so that its documented

nickovaras:
Can do that certainly.

PR #157, opened 2020-01-31
store shared files in share/{PROJECT_NAME} rather than share/

Jconn:
fixes: https://github.com/ros-planning/navigation2/issues/1480


SteveMacenski:
Can you submit one for Dashing as well?

Jconn:
Will do for this and non temporal layer. Didn't even check if there was a dashing branch..my mistake

PR #160, opened 2020-03-05
Add voxel_min_points param for voxel filter

akiroz:
This feature prevents "ghost points" from noise generated by depth cameras.

PR #161, opened 2020-03-05
Add voxel_min_points param for voxel filter (ROS2)

akiroz:
See #160

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ rgbd_obstacle_layer:
     inf_is_valid: false          #default false, for laser scans
     clear_after_reading: true    #default false, clear the buffer after the layer gets readings from it
     voxel_filter: true           #default off, apply voxel filter to sensor, recommend on 
-    voxel_min_points: 10         #default 0, minimum points per voxel for voxel filter
+    voxel_min_points: 0          #default 0, minimum points per voxel for voxel filter
   rgbd1_clear:
     enabled: true                #default true, can be toggled on/off with associated service call
     data_type: PointCloud2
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -116,6 +116,7 @@ rgbd_obstacle_layer:
     inf_is_valid: false          #default false, for laser scans
     clear_after_reading: true    #default false, clear the buffer after the layer gets readings from it
     voxel_filter: true           #default off, apply voxel filter to sensor, recommend on 
+    voxel_min_points: 10         #default 0, minimum points per voxel for voxel filter
```
SteveMacenski: Please change this to `0`.

src/measurement_buffer.cpp
```diff
@@ -154,6 +154,7 @@ void MeasurementBuffer::BufferROSCloud(
       sor.setDownsampleAllData(false);
       float v_s = static_cast<float>(_voxel_size);
       sor.setLeafSize(v_s, v_s, v_s);
+      sor.setMinimumPointsNumberPerVoxel((unsigned int) _voxel_min_points);
```
SteveMacenski: please use `static_cast<unsigned int>(_voxel_min_points);`

SteveMacenski (changes requested):
Same items from ROS1

SteveMacenski (approved):

nickvaras:
There's an issue with minimum points per voxel in PCL, so just wanted to confirm what I'd expect, which is that the feature added in this PR doesn't really have any effect when applied?? Unless it's been fixed in newer versions of PCL...

SteveMacenski:
I'm not sure - I haven't tested this personally, @akiroz I would have thought did since they proposed it

akiroz:
We actually ended up not using this param in the end but I believe it's been fixed in PCL upstream and planned for the coming 1.12 release.

SteveMacenski:
This appears to have potentially broken things per https://github.com/SteveMacenski/spatio_temporal_voxel_layer/issues/187 and https://github.com/SteveMacenski/spatio_temporal_voxel_layer/issues/199

Issue #164, opened 2020-04-06
Is there an error in the minimum-maximum frustum distance?


facontidavide:
Hi,

when I was inspecting the code, I found something quite suspicious.

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/eloquent-devel/src/frustum_models/depth_camera_frustum.cpp#L97-L98

According to this mathematical formula, **min** and **max** distance refer to the distance between the origin and the eight corners of the frustum.

But I think this is NOT what you wanted.

If I don't mistake, **_min_d** and **_max_d** should refer to the distance of the planes parallel to the XY plane and perpendicular to the Z one.

facontidavide:
Right now I am super busy, so I don't have  time to prepare a proper pull request, but I refactored from the ground up the depth camera frustum and the code is much easier to read.

Feel free to use this if you want:

https://bitbucket.org/snippets/df-bor/9nM868/frustum-calculation-fast-and-simple

SteveMacenski:
I think you are correct, it should be  max (or min) * cos(phi/2) by that math, but it is a little convoluted.

facontidavide:
I wanted to use your frustum code, when I realized the very same math could be written in a more simple way 😅

SteveMacenski:
> I wanted to use your frustum code, when I realized the very same math could be written in a more simple way 

Yeah. I wasn't too concerned with the complexity on this since it is cached on startup. Once it worked, it worked, and I moved on. I was also trying to use the openVDB tools since I was learning that at the same time. It could be simplified for sure.

facontidavide:
I love this library, you did a great job... And you game me an excuse to learn OpenVDB 🤗

SteveMacenski:
I barely touched the surface of what its really good at. If you actually set it up as a signed distance field, you can get all the raycasting speeds ups too. However, I really never figured out a way to do that reliably with "real" data from "real" sensors since the surface of obstacles with noisy sensors isnt going to be consistent. Maybe it doesn't matter if I guestimate the surface but the overall the raycasting without the signed distance speed ups weren't enough for my needs. I needed something to drop CPU a bunch so I could process 8+ depth cameras. 

If I were to start from scratch, I'd probably take another crack at it.

facontidavide:
> Yeah. I wasn't too concerned with the complexity on this since it is cached on startup.

[...]

PR #165, opened 2020-04-07
embrace trigonometry (fix also issue #164)

facontidavide:
In **theory**, this modification is just about readability. It should give EXACTLY the same result, but...

In practice, we fixed also a problem with **min/max** distance, as reported in #164 

Please test it before merging it, to be sure I didn't miss anything.

### DIFF
--- a/include/spatio_temporal_voxel_layer/frustum_models/depth_camera_frustum.hpp
+++ b/include/spatio_temporal_voxel_layer/frustum_models/depth_camera_frustum.hpp
@@ -42,6 +42,7 @@
 
 // STVL
 #include <spatio_temporal_voxel_layer/frustum_models/frustum.hpp>
+#include <array>
 
 namespace geometry
 {
@@ -55,7 +56,8 @@ class DepthCameraFrustum : public Frustum
 public:
   DepthCameraFrustum(const double& vFOV, const double& hFOV,
           const double& min_dist, const double& max_dist);
-  virtual ~DepthCameraFrustum(void);
+
+  virtual ~DepthCameraFrustum(void) = default;
 
   // transform plane normals by depth camera pose
   virtual void TransformModel(void);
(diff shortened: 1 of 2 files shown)


facontidavide:
THis is emabrassing.... there is a compilation error. Let me fix it

SteveMacenski:
@tasilb can you test this? If you set `VISUALIZE_FRUSTUM 1` in the depth camera frustum file, it'll publish the points of the frustum (kill performance, but you can see the frustum corners and normal vectors. Just making sure I don't break you guys -- stuff inside updating, stuff outside not, the usual.

tasilb:
Sure thing, though I might have to take my time with it due to hardware availability

________________________________

SteveMacenski:
@tasilb have you taken a look at this?

mlsdpk:
Are there any updates on this? I've noticed a significant improvement in CPU juice after removing the normalize operation

facontidavide:
I guess the changes are still valid 😆

Using std array instead of vector surely help too

SteveMacenski:
@mlsdpk if you can validate that this represents the right frustum now, I can merge this in.

mlsdpk:
Understood! I'll do my best to share results from real hardware, though I cannot guarantee it due to confidentiality concerns. 

Is it acceptable if I submit a PR for unit tests to compare the points from pcl::FrustumCulling filter with STVL's depth camera frustum? This would allow us to thoroughly test the changes introduced in this PR also. Please let me know if this approach aligns with your expectations.

SteveMacenski:
Visualizing the frustum and/or measuring the plane angles and/or measuring the corner points and comparing them to manually computed values from another method would be all ways to validate. 

Certainly pcl would be another way :-) Unit tests are always appreciated, but not required for this repository. This is largely in legacy maintenance mode with few/no outstanding issues (beyond this) that hopefully will be superseded by something building on this idea with probabilistic modeling in the medium term future. 

I usually require much more formality, but this is an old project that predated that policy and I expect that if Davide tested it, it probably works fine, I just want a 3rd opinion since we're all human

PR #170, opened 2020-07-17
dynamic (common) min max obstacle height

nickvaras:
Hi Steve. Congrats by the way on your Best ROS developer nomination. Well deserved.

I'm not sure what to make of this PR. In a way I'm not even sure that it should be merged, but it is certainly meant to spark conversation. It has to do with the ability to dynamically adjust the minimum and maximum obstacle height. The case for those parameters being reconfigurable is, I hope, obvious. There's tons of robots that have some sort means of adjusting height (think mobile manipulators), and conceptually, the fact that we animals can duck under things to not bump into things should make it conceptually evident. 

I do, however, wonder if having the pcl filters built in is backfiring in a way. I get the "batteries included" approach, but at the same time, a nodelet implementation of a voxel or passthorugh filter with zero-copy may be more efficient and is probably in place in most robots anyway, so the benefit of having it built into stvl debatable. Maybe a better PR should be adding the option to skip any pcl filtering altogether for those who want to do it on their own. 

The main issue with this PR is that the added parameter applies to all observation sources. For many robots that have 1 source per layer instance , this may work just fine, but conceptually I don't like the value overriding and ideally each observation source would have its own dynparam server, as we cannot know a-priori the sources so implementing per-source reconfigurable parameters is not trivial (not even know if possible).

node and nodelet implementations of the pcl filters are already dynamically reconfigurable, hence the hesitation here. But, as said, the idea is to at least spark conversation.

### DIFF
--- a/include/spatio_temporal_voxel_layer/measurement_buffer.hpp
+++ b/include/spatio_temporal_voxel_layer/measurement_buffer.hpp
@@ -111,6 +111,8 @@ public:
   // enabler setter getter
   bool IsEnabled(void) const;
   void SetEnabled(const bool& enabled);
+  void SetVerticalLimits(const double& min_obstacle_height,const double& max_obstacle_height);
+
 
   // State knoweldge if sensors are operating as expected
   bool UpdatedAtExpectedRate(void) const;
--- a/include/spatio_temporal_voxel_layer/measurement_buffer.hpp
+++ b/include/spatio_temporal_voxel_layer/measurement_buffer.hpp
@@ -111,7 +111,7 @@ public:
   // enabler setter getter
   bool IsEnabled(void) const;
   void SetEnabled(const bool& enabled);
-  void SetVerticalLimits(const double& min_obstacle_height,const double& max_obstacle_height);
+  void SetVerticalLimits(const double& min_obstacle_height, const double& max_obstacle_height);
 
 
   // State knoweldge if sensors are operating as expected
--- a/src/measurement_buffer.cpp
+++ b/src/measurement_buffer.cpp
@@ -284,6 +284,16 @@ void MeasurementBuffer::SetEnabled(const bool& enabled)
 }
 
 /*****************************************************************************/
+void MeasurementBuffer::SetVerticalLimits(const double& min_obstacle_height,      \
+                                          const double& max_obstacle_height)
+  
+/*****************************************************************************/
+{
+  _min_obstacle_height = min_obstacle_height;
+  _max_obstacle_height = max_obstacle_height;
+}
+
+/*****************************************************************************/
 void MeasurementBuffer::ResetLastUpdatedTime(void)
 /*****************************************************************************/
 {
--- a/src/measurement_buffer.cpp
+++ b/src/measurement_buffer.cpp
@@ -284,7 +284,7 @@ void MeasurementBuffer::SetEnabled(const bool& enabled)
 }
 
 /*****************************************************************************/
-void MeasurementBuffer::SetVerticalLimits(const double& min_obstacle_height,      \
+void MeasurementBuffer::SetVerticalLimits(const double& min_obstacle_height,  \
                                           const double& max_obstacle_height)
   
 /*****************************************************************************/
--- a/src/spatio_temporal_voxel_layer.cpp
+++ b/src/spatio_temporal_voxel_layer.cpp
@@ -597,7 +597,7 @@ void SpatioTemporalVoxelLayer::DynamicReconfigureCallback( \
   _mapping_mode = config.mapping_mode;
   _map_save_duration = ros::Duration(config.map_save_duration);
 
-  if (level >=1) //update grid
+  if (level > 0) //update grid
   {
     auto default_value = (config.track_unknown_space) ? \
                             costmap_2d::NO_INFORMATION : costmap_2d::FREE_SPACE;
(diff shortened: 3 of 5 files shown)


include/spatio_temporal_voxel_layer/measurement_buffer.hpp
```diff
@@ -111,6 +111,8 @@ class MeasurementBuffer
   // enabler setter getter
   bool IsEnabled(void) const;
   void SetEnabled(const bool& enabled);
+  void SetVerticalLimits(const double& min_obstacle_height,const double& max_obstacle_height);
```
SteveMacenski, line 114: space between comma and const please

src/measurement_buffer.cpp
```diff
@@ -283,6 +283,16 @@ void MeasurementBuffer::SetEnabled(const bool& enabled)
   _enabled = enabled;
 }
 
+/*****************************************************************************/
+void MeasurementBuffer::SetVerticalLimits(const double& min_obstacle_height,      \
```
SteveMacenski, line 287: `\` at the end of the `/*****/` please

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -612,6 +612,15 @@ void SpatioTemporalVoxelLayer::DynamicReconfigureCallback( \
       static_cast<double>(default_value_), _decay_model, \
       _voxel_decay, _publish_voxels);
   }
+
+  if (level >1) //Override all observations' vertical range with these common parameters
```
SteveMacenski, line 616: Space between `> 1` please

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -612,6 +612,15 @@ void SpatioTemporalVoxelLayer::DynamicReconfigureCallback( \
...
+  {
+    observation_buffers_iter it = _observation_buffers.begin();
+    for (it; it != _observation_buffers.end(); ++it)
+    {
+      (*it)->SetVerticalLimits(config.general_min_obstacle_height,config.general_max_obstacle_height);
```
SteveMacenski, line 621: Space between comma please

SteveMacenski:
> but at the same time, a nodelet implementation of a voxel or passthorugh filter with zero-copy may be more efficient and is probably in place in most robots anyway, so the benefit of having it built into stvl debatable

You're also welcome to not use them by unsetting them and then having your own components do it - the passthrough is probably necessary but the voxel you're right it could be external and that would be fine too. I did that mostly because for very dense sensors, it lowers CPU dramatically in processing & its easier to flip a switch than to setup a new nodelet in testing to see if it helps. 

Though in no situation would a nodelet be more efficient. Nothing beats direct memory management in sequential order. A nodelet or anything else will always have more instructions and therefore latency. 

I think the full layer reset of these settings could be not great, as you mention, if you have multiple sensors each configured beforehand with different values. I don't have a particular problem with merging this with that though as long as its documented that it will override all sources in the layer no matter what they were set to. You could expose a services for each individual source to change some parameters, but that's a bit of an undertaking and if you did that, you'd have to do all of the relevant parameters.

nickvaras:
Ok. I've made the requested changes and included additional documentation in the README explaining the behavior of the new functionality. 

Just as a thought experiment, regarding 

> Though in no situation would a nodelet be more efficient

what if you have a sensor that outputs a very dense cloud (and has a nodelet compatible driver), that let's you use a zero copy nodelet to voxelize/downsample that cloud before serializing to send over to move_base. I do not know for sure, but at least I can imagine a case where the overhead of serializing the message so send it over is comparable to the benefit of doing the benefit of applying the filter in-node.... just a thought.... I wish the PCL filters were stackable in a single iteration, so all filtering criteria are run in a single iteration. It seems to me that the same sensor data gets iterated multiple times on most sensor pipeline implementations I've seen...

SteveMacenski:
If the sensor driver was a nodelet and the [algorithm] (voxel, etc) was done there rather than in this layer, yes, it would be faster. I'm just talking about stand alone nodelet chains though. It also depends on the filter and if it actually removes data or not.

nickvaras:
If I remember well, both Realsense and Velodyne drivers are run as nodelets.

SteveMacenski:
OK, welcome to use other filtering - there's no requirement to use it here. Its just included for convenience

nickvaras:
> OK, welcome to use other filtering - there's no requirement to use it here. Its just included for convenience

And that's great, just to be clear I'm not complaining. I think the "batteries included" approach is great. But there's no not-using the included filters option as far as I can see, it's either voxel or passthrough, which means iteration on the entire pointcloud message anyway. That's why I suggested that instead of [voxel:true or false] we could have [filter: voxel or passthrough or none]
Maybe I should just comment less and PR more...

```
    if ( _voxel_filter )
    {
      pcl::VoxelGrid<pcl::PCLPointCloud2> sor;
      sor.setInputCloud (cloud_pcl);
      sor.setFilterFieldName("z");
      sor.setFilterLimits(_min_obstacle_height,_max_obstacle_height);
      sor.setDownsampleAllData(false);
      sor.setLeafSize ((float)_voxel_size,
                       (float)_voxel_size,
                       (float)_voxel_size);
      sor.setMinimumPointsNumberPerVoxel(static_cast<unsigned int>(_voxel_min_points));
      sor.filter(*cloud_filtered);
    }
    else
    {
      pcl::PassThrough<pcl::PCLPointCloud2> pass_through_filter;
      pass_through_filter.setInputCloud(cloud_pcl);
      pass_through_filter.setKeepOrganized(false);
      pass_through_filter.setFilterFieldName("z");
      pass_through_filter.setFilterLimits( \
                  _min_obstacle_height,_max_obstacle_height);
      pass_through_filter.filter(*cloud_filtered);
    }
```

SteveMacenski:
Looking at that logic now - I think it should be passthrough regardless so that we remove any data that is clearly not required to be processed. The voxel should be flexible. The reason I wouldn't want to disable passthrough is because then you're making parameters like min/max height and resolution (which are layer parameters) parameters of your filtering pipeline which breaks encapsulation of ideas. Plus, very, very, very few users will even look at this to know that they needed to have pre-filtered this to work properly so removing it would likely end up with the vast majority of users having a bad experience.

nickvaras:
What about defaulting to passthrough as it is currently, but adding the option of skipping filtering for those few users that want to do the filtering on their own. One could argue that the computational cost of passthrough is low, but it is the dynamic reconfigurability that is, probably obvious by now, my main concern. Either way, I think that it is constructive to talk about these things.

nickvaras:
Having second thoughts about this approach. Too clunky.

PR #171, opened 2020-07-18
Filter opt out

nickvaras:
I like this PR much better. It doesn't break API or change the user experience for existing users in any way, it follows the coding conventions already in place, nomenclature, etc. It's clean and light.


src/measurement_buffer.cpp
```diff
@@ -173,9 +176,9 @@ void MeasurementBuffer::BufferROSCloud(const sensor_msgs::PointCloud2& cloud)
       pass_through_filter.setFilterLimits( \
                   _min_obstacle_height,_max_obstacle_height);
       pass_through_filter.filter(*cloud_filtered);
+      pcl_conversions::fromPCL(*cloud_filtered, *cld_global);
```
SteveMacenski, line 179: Seems like you should move the `    pcl_conversions::toPCL(*cld_global, *cloud_pcl);` line into both the if/else statements if no longer requires either. That conversion is non-trivial in time

nickvaras: Sorry if I'm extra slow today, but I'm not sure what you mean. I thought that's what I did. That line only runs if one of the filters is active, else it's skipped. The only `pcl_conversions::fromPCL` in all of STVL seem to be on measurement_buffer.cpp , both in those blocks. Could you please elaborate a bit?

nickvaras: Excellent suggestion!

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -175,6 +176,8 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
     source_node.param("decay_acceleration", decay_acceleration, 0.);
     // performs an approximate voxel filter over the data to reduce
     source_node.param("voxel_filter", voxel_filter, false);
+    // Skips built-in pcl filtering altogether (BYO filtering)
```
SteveMacenski, line 179: Skips? I think this should be describing what the param is / does. 

Please also add this param to the param description in the readme

nickvaras: Good catch. An early implementation was negative, i.e., skip_filtering or something like that.

README.md
```diff
@@ -131,6 +131,7 @@ rgbd_obstacle_layer:
     inf_is_valid: false          #default false, for laser scans
     clear_after_reading: true    #default false, clear the buffer after the layer gets readings from it
     voxel_filter: true           #default off, apply voxel filter to sensor, recommend on 
+    passthrough_filter: true     #default true, crop data by min & max height if voxel_filter off
```
SteveMacenski, line 134: Need some docs about one versus the other and what takes priority if both are set to true (like say, in this file)

SteveMacenski:
the more this PR evolves, the more I think this shouldn't be merged. It creates confusion and some of the descriptions aren't where I would expect them.

I think we should just close this out. If you know this is here, you can remove that from your pre-processing pipeline into STVL. I think the more this change is made, the less logic it makes

nickvaras:
The wording, description of features and their placement, are all subject to personal preferences and I respect your opinion as the author. I've obviously tried to accommodate for your preferences through the multiple change requests, and am not invested in any particular way when it comes to documentation or commenting of the code. If up to me, I'd even leave the new feature hidden for those brave enough to look at the source code. However, when it comes to software functionality, things are pretty black and white. It is a simple fact that the current implementation forces one of two filtering options on the user, and does not allow for dynamic reconfiguration of those parameters. While providing functionality for convenience is welcome by all, forcing that functionality does not add value per se. Confusion can be avoided by other means. Moreover, it takes away the option from the user to use nodelet-based filter implementations that feed directly off the sensor's driver. Again, those are just facts and nor subject to opinion. 
A while ago you starred the Linefit Ground segmentation repo, so I assume you are aware that sensor data is often processed in more ways than just voxelization or cropping before it's used for navigation. I find that suggesting that I modify my pipeline so more of the filtering takes place in STVL, which is only active as a costmap plugin, is rather narrow sighted. I've always thought this is a great library and tried to contribute because I think it's a shame to see great projects get forked and diverge over stuff that at the end of the day is not that important. I've attempted to add something that increases options for the user while taking away nothing and requiring no changes on anybody's part. While I can always just use my own version, I thought that there may be other users out there that may find the new feature useful, and I was happy to share this and future additions with the rest of the community.

SteveMacenski:
My main issue with this PR is in the implementation not in the philosophy behind it. I have on my queue to reimplement what you did in a way that makes me more OK with it, but naively adding 2 parameters without any meaningful description in their creation in the code or yaml file that they are mutually exclusive and the way they override each other are **critical**. 

Further, its obviously unideal to even have 2 separate parameters if they're going to be mutually exclusive. It should either be an enum or string field to describe the filter (or no filter) to use. 

As you can see, there are real design reasons around why this PR shouldn't be merged. I don't appreciate the wall of text. I can appreciate that you're unhappy that I don't feel this should be merged, but as the PR continues to evolve there are clear holes being created.

nickvaras:
> adding 2 parameters 

Only a single parameter is added in this PR

>without any meaningful description in their creation in the code or yaml file that they are mutually exclusive and the way they override each other are **critical**.

Agreed. Very, this is important to convey. The provided description was just a first attempt.
 
> Further, its obviously unideal to even have 2 separate parameters if they're going to be mutually exclusive. It should either be an enum or string field to describe the filter (or no filter) to use.

Absolutely, Indeed that was an early concept to implement, but that could things for existing users. So the PR implementation was a compromise.

> As you can see, there are real design reasons around why this PR shouldn't be merged. I don't appreciate the wall of text. I can appreciate that you're unhappy that I don't feel this should be merged, but as the PR continues to evolve there are clear holes being created. I have always been flexible to add people's needs to projects I support and build, but flexibility doesn't mean merging every PR that is submitted.

Certainly not every PR should be submitted, I don't suggest otherwise. But regarding design reasons/principles, modularity ranks high up in the ROS ecosystem, and forcibly absorbing functionality that already exists in other packages is obviously our main point of disagreement. Let it be said that I still highly appreciate the amazing work you've done and, disappointing as it may be to see this rejected, I still look forward to future improvements in the projects that you lead.

SteveMacenski:
I'll still make the changes you requested, I'm going to just go about it a slightly different way.

PR #172, opened 2020-07-22
Adding optional filter

SteveMacenski:
#171 extension, @nickvaras please review 

This will need to be retargeted to a Noetic branch and merged there, we cannot change API / ABI in the middle of a distribution but it should work on a melodic system if you want.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ rgbd_obstacle_layer:
     observation_persistence: 0.0 #default 0, use all measurements taken during now-value, 0=latest 
     inf_is_valid: false          #default false, for laser scans
     clear_after_reading: true    #default false, clear the buffer after the layer gets readings from it
-    voxel_filter: true           #default off, apply voxel filter to sensor, recommend on 
+    filter: "voxel"              #default passthrough, apply "voxel", "passthrough", or no filter to sensor data, recommend on
     voxel_min_points: 0          #default 0, minimum points per voxel for voxel filter
   rgbd1_clear:
     enabled: true                #default true, can be toggled on/off with associated service call
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ Required dependencies ROS Kinetic, navigation, OpenVDB, TBB.
 
 An example fully-described configuration is shown below.
 
-Note: We supply 2 PCL filters within STVL to use to massage the data to lower compute overhead. STVL has an approximate voxel filter to make the data more sparse if very dense. It also has a passthrough filter to limit processing data within the valid minimum to maximum height bounds. The voxel filter is recommended if it lowers CPU overhead, otherwise, passthrough filter. No filter is also available if you pre-process your data or are not interested in performance optimizations.
+Note: We supply two PCL filters within STVL to massage the data to lower compute overhead. STVL has an approximate voxel filter to make the data more sparse if very dense. It also has a passthrough filter to limit processing data within the valid minimum to maximum height bounds. The voxel filter is recommended if it lowers CPU overhead, otherwise, passthrough filter. No filter is also available if you pre-process your data or are not interested in performance optimizations.
 
 ```
 rgbd_obstacle_layer:
--- a/README.md
+++ b/README.md
@@ -99,6 +99,8 @@ Required dependencies ROS Kinetic, navigation, OpenVDB, TBB.
 
 An example fully-described configuration is shown below.
 
+Note: We supply 2 PCL filters within STVL to use to massage the data to lower compute overhead. STVL has an approximate voxel filter to make the data more sparse if very dense. It also has a passthrough filter to limit processing data within the valid minimum to maximum height bounds. The voxel filter is recommended if it lowers CPU overhead, otherwise, passthrough filter. No filter is also available if you pre-process your data or are not interested in performance optimizations.
+
 ```
 rgbd_obstacle_layer:
   enabled:               true
@@ -130,7 +132,7 @@ rgbd_obstacle_layer:
     observation_persistence: 0.0 #default 0, use all measurements taken during now-value, 0=latest 
     inf_is_valid: false          #default false, for laser scans
     clear_after_reading: true    #default false, clear the buffer after the layer gets readings from it
-    filter: "voxel"              #default passthrough, apply "voxel", "passthrough", or no filter to sensor data, recommend on
+    filter: "voxel"              #default passthrough, apply "voxel", "passthrough", or no filter to sensor data, recommended to have at one filter on
     voxel_min_points: 0          #default 0, minimum points per voxel for voxel filter
   rgbd1_clear:
     enabled: true                #default true, can be toggled on/off with associated service call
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -130,7 +130,7 @@ rgbd_obstacle_layer:
     observation_persistence: 0.0 #default 0, use all measurements taken during now-value, 0=latest 
     inf_is_valid: false          #default false, for laser scans
     clear_after_reading: true    #default false, clear the buffer after the layer gets readings from it
-    voxel_filter: true           #default off, apply voxel filter to sensor, recommend on 
+    filter: "voxel"              #default passthrough, apply "voxel", "passthrough", or no filter to sensor data, recommend on
```
nickvaras, line 133:
```suggestion
    filter: "voxel"              #default passthrough, apply "voxel", "passthrough", or no filter to sensor data, recommend voxel```

File: example/constrained_indoor_environment_config.yaml [hunk 2]
```diff
@@ -26,7 +26,7 @@ rgbd_obstacle_layer:
     expected_update_rate: 0.0    # default 0, if not updating at this rate at least, remove from buffer
     observation_persistence: 0.0 # default 0, use all measurements taken during now-value, 0=latest
     inf_is_valid: false          # default false, for laser scans
-    voxel_filter: false          # default off, apply voxel filter to sensor, recommend on
+    filter: "passthrough"        # default passthrough, apply "voxel", "passthrough", or no filter to sensor data, recommend on
```
nickvaras, line 29:
```suggestion
    filter: "passthrough"        # default passthrough, apply "voxel", "passthrough", or no filter to sensor data, recommend voxel
```

example/outdoor_environment_config.yaml
```diff
@@ -26,7 +26,7 @@ rgbd_obstacle_layer:
     expected_update_rate: 0.0    # default 0, if not updating at this rate at least, remove from buffer
     observation_persistence: 0.0 # default 0, use all measurements taken during now-value, 0=latest
     inf_is_valid: false          # default false, for laser scans
-    voxel_filter: false          # default off, apply voxel filter to sensor, recommend on
+    filter: "passthrough"        # default passthrough, apply "voxel", "passthrough", or no filter to sensor data, recommend on
```
nickvaras, line 29:
```suggestion
    filter: "passthrough"        # default passthrough, apply "voxel", "passthrough", or no filter to sensor data, recommend voxel
```

example/standard_indoor_environment_config.yaml
```diff
@@ -26,7 +26,7 @@ rgbd_obstacle_layer:
     expected_update_rate: 0.0    # default 0, if not updating at this rate at least, remove from buffer
     observation_persistence: 0.0 # default 0, use all measurements taken during now-value, 0=latest
     inf_is_valid: false          # default false, for laser scans
-    voxel_filter: false          # default off, apply voxel filter to sensor, recommend on
+    filter: "passthrough"        # default passthrough, apply "voxel", "passthrough", or no filter to sensor data, recommend on
```
nickvaras, line 29:
```suggestion
    filter: "passthrough"        # default passthrough, apply "voxel", "passthrough", or no filter to sensor data, recommend voxel
```

example/vlp16_config.yaml
```diff
@@ -26,7 +26,7 @@ rgbd_obstacle_layer:
     expected_update_rate: 0.0    # default 0, if not updating at this rate at least, remove from buffer
     observation_persistence: 0.0 # default 0, use all measurements taken during now-value, 0=latest
     inf_is_valid: false          # default false, for laser scans
-    voxel_filter: false          # default off, apply voxel filter to sensor, recommend on
+    filter: "voxel"              # default passthrough, apply "voxel", "passthrough", or no filter to sensor data, recommend on
```
nickvaras, line 29:
```suggestion
    filter: "voxel"              # default passthrough, apply "voxel", "passthrough", or no filter to sensor data, recommend voxel
```

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -173,8 +174,8 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
     source_node.param("horizontal_fov_angle", hFOV, 1.04);
     // acceleration scales the model's decay in presence of readings
     source_node.param("decay_acceleration", decay_acceleration, 0.);
-    // performs an approximate voxel filter over the data to reduce
-    source_node.param("voxel_filter", voxel_filter, false);
+    // performs a PCL filter, either an approximate voxel filter or passthrough filter
```
nickvaras, line 177:
```suggestion
    // Apply a PCL filter (Approximate VoxeGrid or PassThrough) or skip
```

nickvaras:
This is definitely the right way to implement this when breaking API is an option. Thanks for implementing this and sorry about any displeasure this may have caused.

SteveMacenski:
Well I think the point was to have recommended something on, I don't recommend one vs the other generically, it depends on the sensor and resolution. I didn't want to bias one over another because they're both valid for different situations

nickvaras:
> Well I think the point was to have recommended something on, I don't recommend one vs the other generically, it depends on the sensor and resolution. I didn't want to bias one over another because they're both valid for different situations

Oh, I thought it had just been overlooked and left there from the previous description. If the parameter has three options, and none of them is "on", then recommending "on" may be confusing. The recommendation should really be that downsampling/decimation and vertical cropping of sensor data is definitely recommended, either by using the built-in convenience filters or before the data is passed on to STVL

SteveMacenski:
How's that?

SteveMacenski:
OK now?

doisyg:
This new way of using filters is not very clear, I got tricked after trying it. Let me explain:
In my current workflow, I use an external down-sampling filter but I rely on STVL for vertical cropping according to the `min_obstacle_height` and `max_obstacle_height` params. So I use to have `voxel_filter: false `
With this PR, I set `filter: ""` and expected to have vertical cropping applied, but I understand now that I should set `filter: "passthrough"`. In my opinion a warning should be displayed in the console if `min_obstacle_height` and `max_obstacle_height` are set and won't be used.

Additionaly, this is not my use case, but I believe that when setting `filter: "voxel"` , `min_obstacle_height` and `max_obstacle_height` are also not applied, but before when setting `voxel_filter: true `, there were. It seems there is no way anymore of using both downsampling and vertical cropping in the same time. (but again, not my use-case)

SteveMacenski:
>  In my opinion a warning should be displayed in the console if min_obstacle_height and max_obstacle_height are set and won't be used.

I'd generally agree if they're non-default parameters, but most people I see are just copy pasting my example config files (what I made them for) so in gain tuning, I don't want to be throwing out errors like that. But I understand the sentiment, I would merge a PR do implement that. 

> I believe that when setting filter: "voxel" , min_obstacle_height and max_obstacle_height are also not applied, but before when setting voxel_filter: true , there were

The actual logic of the voxel filter once it gets to it hasn't changed https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/noetic-devel/src/measurement_buffer.cpp#L158 it was just the `if` condition criteria that was. https://github.com/SteveMacenski/spatio_temporal_voxel_layer/pull/172/files#diff-f66af5a8d1e47c539d2a0b39d86d43aaL159 - so I don't see any changes made in that area to cause an issue unless you're just not seeing the voxel filter _at all_ working.

doisyg:
> I'd generally agree if they're non-default parameters, but most people I see are just copy pasting my example config files (what I made them for) so in gain tuning, I don't want to be throwing out errors like that. But I understand the sentiment, I would merge a PR do implement that.

I am fine with no warnings now that I noticed the change ^^ (but one using an old config file could get tricked too)



> Can you try that again and let me know what you think?

It works ::facepalm::, I misread the commit diff

PR #180, opened 2020-09-18
FIX-179 making STVL compatible with nav2 main branch

garyrightbot:
Build is successful though I am yet to test it.

### DIFF
--- a/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
+++ b/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
@@ -173,6 +173,8 @@ private:
   std::vector<observation::MeasurementReading> _static_observations;
   std::unique_ptr<volume_grid::SpatioTemporalVoxelGrid> _voxel_grid;
   boost::recursive_mutex _voxel_grid_lock;
+  rclcpp::Clock::SharedPtr clock;
+  rclcpp::Logger logger{rclcpp::get_logger("spatio_temporal_voxel_layer")};
 };
 
 }  // namespace spatio_temporal_voxel_layer
--- a/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
+++ b/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
@@ -173,7 +173,7 @@ private:
   std::vector<observation::MeasurementReading> _static_observations;
   std::unique_ptr<volume_grid::SpatioTemporalVoxelGrid> _voxel_grid;
   boost::recursive_mutex _voxel_grid_lock;
-  rclcpp::Clock::SharedPtr clock;
+  rclcpp::Clock::SharedPtr _clock;
   rclcpp::Logger logger{rclcpp::get_logger("spatio_temporal_voxel_layer")};
 };
 
--- a/src/spatio_temporal_voxel_layer.cpp
+++ b/src/spatio_temporal_voxel_layer.cpp
@@ -66,12 +66,22 @@ SpatioTemporalVoxelLayer::~SpatioTemporalVoxelLayer(void)
 void SpatioTemporalVoxelLayer::onInitialize(void)
 /*****************************************************************************/
 {
-  RCLCPP_INFO((node_.lock())->get_logger(),
+
+  auto node = node_.lock();
+  
+  if (!node) {
+    throw std::runtime_error{"Failed to lock node"};
+  }
+  
+  clock = node->get_clock();
+  logger = node->get_logger();
+
+  RCLCPP_INFO(logger,
     "%s being initialized as SpatioTemporalVoxelLayer!", getName().c_str());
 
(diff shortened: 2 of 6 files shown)


src/spatio_temporal_voxel_layer.cpp
```diff
@@ -66,12 +66,12 @@ SpatioTemporalVoxelLayer::~SpatioTemporalVoxelLayer(void)
 void SpatioTemporalVoxelLayer::onInitialize(void)
 /*****************************************************************************/
 {
-  RCLCPP_INFO(node_->get_logger(),
+  RCLCPP_INFO((node_.lock())->get_logger(),
```
SteveMacenski, line 69: At the start of this function you should `auto node = node_lock()` and check if its valid rather than continuously locking the weak pointer, that's not good behavior.

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -513,7 +513,7 @@ void SpatioTemporalVoxelLayer::activate(void)
 /*****************************************************************************/
 {
   // subscribe and place info in buffers from sensor sources
-  RCLCPP_INFO(node_->get_logger(), "%s was activated.", getName().c_str());
+  RCLCPP_INFO((node_.lock())->get_logger(), "%s was activated.", getName().c_str());
```
SteveMacenski, line 516: in `onInitialize` you should store the logger object while the shared pointer is locked as a class member and change all these log statements to using that logger object

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -695,12 +695,12 @@ void SpatioTemporalVoxelLayer::updateBounds(
   // navigation mode: clear observations, mapping mode: save maps and publish
   bool should_save = false;
   if (_map_save_duration) {
-    should_save = node_->now() - _last_map_save_time > *_map_save_duration;
+    should_save = (node_.lock())->now() - _last_map_save_time > *_map_save_duration;
```
SteveMacenski, line 698: As well, store a clock object for use

SteveMacenski (changes requested):
Good first run at it, lets just make this more efficient, this could cause deadlocks

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
```diff
@@ -173,6 +173,8 @@ class SpatioTemporalVoxelLayer : public nav2_costmap_2d::CostmapLayer
   std::vector<observation::MeasurementReading> _static_observations;
   std::unique_ptr<volume_grid::SpatioTemporalVoxelGrid> _voxel_grid;
   boost::recursive_mutex _voxel_grid_lock;
+  rclcpp::Clock::SharedPtr clock;
```
SteveMacenski, line 176: These are class variables and should start with `_`.

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
```diff
@@ -173,6 +173,8 @@ class SpatioTemporalVoxelLayer : public nav2_costmap_2d::CostmapLayer
...
+  rclcpp::Logger logger{rclcpp::get_logger("spatio_temporal_voxel_layer")};
```
SteveMacenski, line 177: Don't initialize here

SteveMacenski (changes requested):

SteveMacenski:
What do you mean its the driver? I'm not sure how that could be possible.

SteveMacenski:
In what specific way does it not work?

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
garyrightbot: Okay, will do.

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
garyrightbot: Are you referring to the `logger` or both `logger` and `clock`?

SteveMacenski:
But _why_ was that happening as a result of these ROS2 wrappers?

SteveMacenski:
OK, if you look at a message from both, it should be pretty clear. If you use the `--no-array` you can just see the headers / content without all the console spam. Maybe a timing issue like you mention, maybe a frame transformation issue.

include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
SteveMacenski: `logger` is the one you initialized, so remove that. Also logger is missing the `_`.

SteveMacenski:
@garyrightbot any update? There is just a couple really small things before we can merge this in.

SteveMacenski:
Superseded by #191

PR #196, opened 2021-03-19
fix voxel_decay declared type

doisyg:
With last rolling nightly (I guess https://github.com/ros2/rclcpp/pull/1522) we have the following error at runtime:
```
[controller_server-7] [ERROR] []: Caught exception in callback for transition 10
[controller_server-7] [ERROR] []: Original error: parameter 'stvl_layer.voxel_decay' has invalid type: Wrong parameter type, parameter {stvl_layer.voxel_decay} is of type {integer}, setting it to {double} is not allowed.
```
`voxel_decay` parameter declared type was improperly infered. Fixed


SteveMacenski (approved):

SteveMacenski:
Weird, the config files all have doubles already. Thanks for the fix

PR #200, opened 2021-04-10
fixes #199

### LINKED ISSUE #199: Obstacles are cleared from the voxel layer but not from the local costmap
Edit: forgot to mention, I'm using ros melodic. STVL was installed by `sudo apt-get install ros-melodic-spatio-temporal-voxel-layer`

Hello, I'm having a problem where the obstacles are cleared from the voxel layer but not from the local costmap. **However, this only affects the local costmap, the global costmap works fine!**

You can see [here](https://streamable.com/ulgjql) a short clip I recorded of it happening. The green lines you see once I give it a new nav goal are from the [costmap_converter](http://wiki.ros.org/costmap_converter), proof that the local costmap was indeed not cleared and it's not just a visual bug.

Below are my costmap params:

```yaml
# common_costmap.yaml

robot_base_frame: base_footprint
footprint: [[-0.7, -0.55], [0.7, -0.55], [0.7, 0.55], [-0.7, 0.55]]
publish_frequency: 30

rgdb_obstacles:
  enabled:                  true
  voxel_decay:              30
  decay_model:              0
  voxel_size:               0.05
  track_unknown_space:      true
  max_obstacle_height:      3.0
  unknown_threshold:        15
  mark_threshold:           0
  update_footprint_enabled: true
  combination_method:       1
  obstacle_range:           8.5
  origin_z:                 0.0
  publish_voxel_map:        true
  observation_sources:      realsense_mark realsense_clear

  realsense_mark:
    data_type: PointCloud2
    topic: realsense/depth/points
    marking: true
    clearing: false
    min_obstacle_height: 0.01
    max_obstacle_height: 3.0
    voxel_filter: true
    
  realsense_clear:
    data_type: PointCloud2
    topic: realsense/depth/points
    marking: false
    clearing: true
    min_z: 0.01
    max_z: 9.0
    vertical_fov_angle: 0.96
    horizontal_fov_angle: 1.22
    decay_acceleration: 5
    model_type: 0
```

```yaml
# global_costmap.yaml

global_costmap:
  plugins:
    - {name: map, type: 'costmap_2d::StaticLayer'}
    - {name: rgdb_obstacles, type: 'spatio_temporal_voxel_layer/SpatioTemporalVoxelLayer'}
    - {name: inflation, type: 'costmap_2d::InflationLayer'}

  map:
    subscribe_to_updates: true

  inflation:
    inflation_radius: 1
    cost_scaling_factor: 5
```

```yaml
# local_costmap.yaml

local_costmap:
  plugins:
    - {name: rgdb_obstacles, type: 'spatio_temporal_voxel_layer/SpatioTemporalVoxelLayer'}

  rolling_window: true
  width: 6
  height: 6
  resolution: 0.05
```

siferati:
Implementation of the [suggestion](https://github.com/SteveMacenski/spatio_temporal_voxel_layer/issues/199#issuecomment-816915693) by @SteveMacenski to fix #199:

> The fix I could see would be to touch the cells cleared as they were cleared to resize the bounds

Cells that were cleared are stored in an `unordered_set` and touched when updating the ROS costmap.

Bugs fixed in this PR:

- [x] Cleared cells are not touched, so the corresponding area of the costmap isn't updated. The `cleared_cells` container solves this.
- [x] Cleared cells are still marked in the cosmap as obstacles. The `cleared_point` flag solves this.

(11/05) Edit: Updated to better describe the work that was done.

### DIFF
--- a/src/spatio_temporal_voxel_grid.cpp
+++ b/src/spatio_temporal_voxel_grid.cpp
@@ -301,7 +301,7 @@ void SpatioTemporalVoxelGrid::Mark(const \
 
 /*****************************************************************************/
 void SpatioTemporalVoxelGrid::operator()(const \
-                                    observation::MeasurementReading& obs)
+                                    observation::MeasurementReading& obs) const
 /*****************************************************************************/
 {
   if (obs._marking)
@@ -328,10 +328,8 @@ void SpatioTemporalVoxelGrid::operator()(const \
       openvdb::Vec3d mark_grid(this->WorldToIndex( \
                                  openvdb::Vec3d(*iter_x, *iter_y, *iter_z)));
 
-      openvdb::Coord pt_index(mark_grid[0], mark_grid[1], mark_grid[2]);
-      PopulateCostmapAndPointcloud(pt_index);
-      
-      if(!this->MarkGridPoint(pt_index, cur_time))
+      if(!this->MarkGridPoint(openvdb::Coord(mark_grid[0], mark_grid[1], \
+                                             mark_grid[2]), cur_time))
       {
         std::cout << "Failed to mark point." << std::endl;
       }
--- a/src/spatio_temporal_voxel_grid.cpp
+++ b/src/spatio_temporal_voxel_grid.cpp
@@ -100,7 +100,7 @@ void SpatioTemporalVoxelGrid::InitializeGrid(void)
 /*****************************************************************************/
 void SpatioTemporalVoxelGrid::ClearFrustums( \
       const std::vector<observation::MeasurementReading>& clearing_readings, \
-      std::unordered_set<occupany_cell>& updated_cells)
+      std::unordered_set<occupany_cell>& cleared_cells)
 /*****************************************************************************/
 {
   boost::unique_lock<boost::mutex> lock(_grid_lock);
@@ -118,7 +118,7 @@ void SpatioTemporalVoxelGrid::ClearFrustums( \
 
   if(clearing_readings.size() == 0)
   {
-    TemporalClearAndGenerateCostmap(obs_frustums, updated_cells);
+    TemporalClearAndGenerateCostmap(obs_frustums, cleared_cells);
     return;
   }
 
@@ -157,14 +157,14 @@ void SpatioTemporalVoxelGrid::ClearFrustums( \
     frustum->TransformModel();
     obs_frustums.emplace_back(frustum, it->_decay_acceleration);
   }
-  TemporalClearAndGenerateCostmap(obs_frustums, updated_cells);
+  TemporalClearAndGenerateCostmap(obs_frustums, cleared_cells);
   return;
 }
 
 /*****************************************************************************/
 void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
                               std::vector<frustum_model>& frustums,           \
-                              std::unordered_set<occupany_cell>& updated_cells)
+                              std::unordered_set<occupany_cell>& cleared_cells)
 /*****************************************************************************/
 {
   // sample time once for all clearing readings
@@ -175,7 +175,7 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
   for (cit_grid; cit_grid.test(); ++cit_grid)
   {
     const openvdb::Coord pt_index(cit_grid.getCoord());
-    const openvdb::Vec3d pose_world = _grid->indexToWorld(pt_index);
+    const openvdb::Vec3d pose_world = this->IndexToWorld(pt_index);
 
     std::vector<frustum_model>::iterator frustum_it = frustums.begin();
     bool frustum_cycle = false;
@@ -203,13 +203,10 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
         if (time_until_decay < 0.)
         {
           // expired by acceleration
+          cleared_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
           if(!this->ClearGridPoint(pt_index))
           {
             std::cout << "Failed to clear point." << std::endl;
-          }
-          else
-          {
-            updated_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
           }
         }
         else
@@ -231,20 +228,16 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
       if (base_duration_to_decay < 0.)
       {
         // expired by temporal clearing
+        cleared_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
         if(!this->ClearGridPoint(pt_index))
         {
           std::cout << "Failed to clear point." << std::endl;
-        }
-        else
-        {
-          updated_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
         }
         continue;
       }
     }
     // if here, we can add to costmap and PC2
     PopulateCostmapAndPointcloud(pt_index);
-    updated_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
   }
 }
 
@@ -254,7 +247,7 @@ void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(const \
 /*****************************************************************************/
 {
   // add pt to the pointcloud and costmap
-  openvdb::Vec3d pose_world = _grid->indexToWorld(pt);
+  openvdb::Vec3d pose_world = this->IndexToWorld(pt);
 
   if (_pub_voxels)
   {
@@ -279,6 +272,33 @@ void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(const \
 }
 
 /*****************************************************************************/
+void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(const \
+                                                    openvdb::Vec3d& pose_world)
+/*****************************************************************************/
+{
+  std::unordered_map<occupany_cell, uint>::iterator cell;
+  cell = _cost_map->find(occupany_cell(pose_world[0], pose_world[1]));
+  if (cell != _cost_map->end())
+  {
+    cell->second += 1;
+  }
+  else
+  {
+    _cost_map->insert(std::make_pair( \
+                              occupany_cell(pose_world[0], pose_world[1]), 1));
+
+    if (_pub_voxels)
+    {
+      geometry_msgs::Point32 point;
+      point.x = pose_world[0];
+      point.y = pose_world[1];
+      point.z = pose_world[2];
+      _grid_points->push_back(point);
+    }
+  }
+}
+
+/*****************************************************************************/
 void SpatioTemporalVoxelGrid::Mark(const \
                 std::vector<observation::MeasurementReading>& marking_readings)
 /*****************************************************************************/
@@ -299,7 +319,7 @@ void SpatioTemporalVoxelGrid::Mark(const \
 
 /*****************************************************************************/
 void SpatioTemporalVoxelGrid::operator()(const \
-                                    observation::MeasurementReading& obs) const
+                                    observation::MeasurementReading& obs)
 /*****************************************************************************/
 {
   if (obs._marking)
@@ -323,8 +343,9 @@ void SpatioTemporalVoxelGrid::operator()(const \
       {
         continue;
       }
-      openvdb::Vec3d mark_grid(this->WorldToIndex( \
-                                 openvdb::Vec3d(*iter_x, *iter_y, *iter_z)));
+      openvdb::Vec3d pose_world(*iter_x, *iter_y, *iter_z);      
+      PopulateCostmapAndPointcloud(pose_world);
+      openvdb::Vec3d mark_grid(this->WorldToIndex(pose_world));
 
       if(!this->MarkGridPoint(openvdb::Coord(mark_grid[0], mark_grid[1], \
                                              mark_grid[2]), cur_time))
--- a/src/spatio_temporal_voxel_grid.cpp
+++ b/src/spatio_temporal_voxel_grid.cpp
@@ -205,7 +205,6 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
         {
           // expired by acceleration
           cleared_point = true;
-          cleared_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
           if(!this->ClearGridPoint(pt_index))
           {
             std::cout << "Failed to clear point." << std::endl;
@@ -231,16 +230,20 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
       {
         // expired by temporal clearing
         cleared_point = true;
-        cleared_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
         if(!this->ClearGridPoint(pt_index))
         {
           std::cout << "Failed to clear point." << std::endl;
         }
       }
     }
-    // if here, we can add to costmap and PC2
-    if (!cleared_point)
+    
+    if (cleared_point)
     {
+      cleared_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
+    }
+    else
+    {
+      // if here, we can add to costmap and PC2
       PopulateCostmapAndPointcloud(pt_index, true);
     }
   }
--- a/src/spatio_temporal_voxel_grid.cpp
+++ b/src/spatio_temporal_voxel_grid.cpp
@@ -209,6 +209,7 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
           {
             std::cout << "Failed to clear point." << std::endl;
           }
+          break;
         }
         else
         {
@@ -244,48 +245,38 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
     else
     {
       // if here, we can add to costmap and PC2
-      PopulateCostmapAndPointcloud(pt_index, true);
+      PopulateCostmapAndPointcloud(pt_index);
     }
   }
 }
 
 /*****************************************************************************/
-void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(                   \
-                                                   const openvdb::Coord& pt,  \
-                                                   const bool& pub_every_voxel)
+void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(const \
+                                                            openvdb::Coord& pt)
 /*****************************************************************************/
 {
   // add pt to the pointcloud and costmap
-  openvdb::Vec3d pose_world = this->IndexToWorld(pt);
+  openvdb::Vec3d pose_world = _grid->indexToWorld(pt);
+
+  if (_pub_voxels)
+  {
+    geometry_msgs::Point32 point;
+    point.x = pose_world[0];
+    point.y = pose_world[1];
+    point.z = pose_world[2];
+    _grid_points->push_back(point);
+  }
 
   std::unordered_map<occupany_cell, uint>::iterator cell;
   cell = _cost_map->find(occupany_cell(pose_world[0], pose_world[1]));
   if (cell != _cost_map->end())
   {
     cell->second += 1;
-
-    if (_pub_voxels && pub_every_voxel)
-    {
-      geometry_msgs::Point32 point;
-      point.x = pose_world[0];
-      point.y = pose_world[1];
-      point.z = pose_world[2];
-      _grid_points->push_back(point);
-    }
   }
   else
   {
     _cost_map->insert(std::make_pair( \
                               occupany_cell(pose_world[0], pose_world[1]), 1));
-
-    if (_pub_voxels)
-    {
-      geometry_msgs::Point32 point;
-      point.x = pose_world[0];
-      point.y = pose_world[1];
-      point.z = pose_world[2];
-      _grid_points->push_back(point);
-    }
   }
 }
 
@@ -338,7 +329,7 @@ void SpatioTemporalVoxelGrid::operator()(const \
                                  openvdb::Vec3d(*iter_x, *iter_y, *iter_z)));
 
       openvdb::Coord pt_index(mark_grid[0], mark_grid[1], mark_grid[2]);
-      PopulateCostmapAndPointcloud(pt_index, false);
+      PopulateCostmapAndPointcloud(pt_index);
       
       if(!this->MarkGridPoint(pt_index, cur_time))
       {
--- a/src/spatio_temporal_voxel_grid.cpp
+++ b/src/spatio_temporal_voxel_grid.cpp
@@ -250,27 +250,27 @@ void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(                   \
   // add pt to the pointcloud and costmap
   openvdb::Vec3d pose_world = this->IndexToWorld(pt);
 
-  if (_pub_voxels && pub_every_voxel)
-  {
-    geometry_msgs::Point32 point;
-    point.x = pose_world[0];
-    point.y = pose_world[1];
-    point.z = pose_world[2];
-    _grid_points->push_back(point);
-  }
-
   std::unordered_map<occupany_cell, uint>::iterator cell;
   cell = _cost_map->find(occupany_cell(pose_world[0], pose_world[1]));
   if (cell != _cost_map->end())
   {
     cell->second += 1;
+
+    if (_pub_voxels && pub_every_voxel)
+    {
+      geometry_msgs::Point32 point;
+      point.x = pose_world[0];
+      point.y = pose_world[1];
+      point.z = pose_world[2];
+      _grid_points->push_back(point);
+    }
   }
   else
   {
     _cost_map->insert(std::make_pair( \
                               occupany_cell(pose_world[0], pose_world[1]), 1));
 
-    if (_pub_voxels && !pub_every_voxel)
+    if (_pub_voxels)
     {
       geometry_msgs::Point32 point;
       point.x = pose_world[0];
--- a/src/spatio_temporal_voxel_grid.cpp
+++ b/src/spatio_temporal_voxel_grid.cpp
@@ -237,19 +237,20 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
       }
     }
     // if here, we can add to costmap and PC2
-    PopulateCostmapAndPointcloud(pt_index);
+    PopulateCostmapAndPointcloud(pt_index, true);
   }
 }
 
 /*****************************************************************************/
-void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(const \
-                                                            openvdb::Coord& pt)
+void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(                   \
+                                                   const openvdb::Coord& pt,  \
+                                                   const bool& pub_every_voxel)
 /*****************************************************************************/
 {
   // add pt to the pointcloud and costmap
   openvdb::Vec3d pose_world = this->IndexToWorld(pt);
 
-  if (_pub_voxels)
+  if (_pub_voxels && pub_every_voxel)
   {
     geometry_msgs::Point32 point;
     point.x = pose_world[0];
@@ -268,26 +269,8 @@ void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(const \
   {
     _cost_map->insert(std::make_pair( \
                               occupany_cell(pose_world[0], pose_world[1]), 1));
-  }
-}
 
-/*****************************************************************************/
-void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(const \
-                                                    openvdb::Vec3d& pose_world)
-/*****************************************************************************/
-{
-  std::unordered_map<occupany_cell, uint>::iterator cell;
-  cell = _cost_map->find(occupany_cell(pose_world[0], pose_world[1]));
-  if (cell != _cost_map->end())
-  {
-    cell->second += 1;
-  }
-  else
-  {
-    _cost_map->insert(std::make_pair( \
-                              occupany_cell(pose_world[0], pose_world[1]), 1));
-
-    if (_pub_voxels)
+    if (_pub_voxels && !pub_every_voxel)
     {
       geometry_msgs::Point32 point;
       point.x = pose_world[0];
@@ -343,12 +326,13 @@ void SpatioTemporalVoxelGrid::operator()(const \
       {
         continue;
       }
-      openvdb::Vec3d pose_world(*iter_x, *iter_y, *iter_z);      
-      PopulateCostmapAndPointcloud(pose_world);
-      openvdb::Vec3d mark_grid(this->WorldToIndex(pose_world));
+      openvdb::Vec3d mark_grid(this->WorldToIndex( \
+                                 openvdb::Vec3d(*iter_x, *iter_y, *iter_z)));
 
-      if(!this->MarkGridPoint(openvdb::Coord(mark_grid[0], mark_grid[1], \
-                                             mark_grid[2]), cur_time))
+      openvdb::Coord pt_index(mark_grid[0], mark_grid[1], mark_grid[2]);
+      PopulateCostmapAndPointcloud(pt_index, false);
+      
+      if(!this->MarkGridPoint(pt_index, cur_time))
       {
         std::cout << "Failed to mark point." << std::endl;
       }
--- a/src/spatio_temporal_voxel_grid.cpp
+++ b/src/spatio_temporal_voxel_grid.cpp
@@ -179,6 +179,7 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
 
     std::vector<frustum_model>::iterator frustum_it = frustums.begin();
     bool frustum_cycle = false;
+    bool cleared_point = false;
 
     const double time_since_marking = cur_time - cit_grid.getValue();
     const double base_duration_to_decay = GetTemporalClearingDuration( \
@@ -203,6 +204,7 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
         if (time_until_decay < 0.)
         {
           // expired by acceleration
+          cleared_point = true;
           cleared_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
           if(!this->ClearGridPoint(pt_index))
           {
@@ -228,16 +230,19 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
       if (base_duration_to_decay < 0.)
       {
         // expired by temporal clearing
+        cleared_point = true;
         cleared_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
         if(!this->ClearGridPoint(pt_index))
         {
           std::cout << "Failed to clear point." << std::endl;
         }
-        continue;
       }
     }
     // if here, we can add to costmap and PC2
-    PopulateCostmapAndPointcloud(pt_index, true);
+    if (!cleared_point)
+    {
+      PopulateCostmapAndPointcloud(pt_index, true);
+    }
   }
 }
 
--- a/src/spatio_temporal_voxel_grid.cpp
+++ b/src/spatio_temporal_voxel_grid.cpp
@@ -98,8 +98,9 @@ void SpatioTemporalVoxelGrid::InitializeGrid(void)
 }
 
 /*****************************************************************************/
-void SpatioTemporalVoxelGrid::ClearFrustums(const \
-               std::vector<observation::MeasurementReading>& clearing_readings)
+void SpatioTemporalVoxelGrid::ClearFrustums( \
+      const std::vector<observation::MeasurementReading>& clearing_readings, \
+      std::unordered_set<occupany_cell>& updated_cells)
 /*****************************************************************************/
 {
   boost::unique_lock<boost::mutex> lock(_grid_lock);
@@ -117,7 +118,7 @@ void SpatioTemporalVoxelGrid::ClearFrustums(const \
 
   if(clearing_readings.size() == 0)
   {
-    TemporalClearAndGenerateCostmap(obs_frustums);
+    TemporalClearAndGenerateCostmap(obs_frustums, updated_cells);
     return;
   }
 
@@ -156,13 +157,14 @@ void SpatioTemporalVoxelGrid::ClearFrustums(const \
     frustum->TransformModel();
     obs_frustums.emplace_back(frustum, it->_decay_acceleration);
   }
-  TemporalClearAndGenerateCostmap(obs_frustums);
+  TemporalClearAndGenerateCostmap(obs_frustums, updated_cells);
   return;
 }
 
 /*****************************************************************************/
 void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
-                                          std::vector<frustum_model>& frustums)
+                              std::vector<frustum_model>& frustums,           \
+                              std::unordered_set<occupany_cell>& updated_cells)
 /*****************************************************************************/
 {
   // sample time once for all clearing readings
@@ -173,6 +175,7 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
   for (cit_grid; cit_grid.test(); ++cit_grid)
   {
     const openvdb::Coord pt_index(cit_grid.getCoord());
+    const openvdb::Vec3d pose_world = _grid->indexToWorld(pt_index);
 
     std::vector<frustum_model>::iterator frustum_it = frustums.begin();
     bool frustum_cycle = false;
@@ -188,7 +191,7 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
         continue;
       }
 
-      if ( frustum_it->frustum->IsInside(this->IndexToWorld(pt_index)) )
+      if ( frustum_it->frustum->IsInside(pose_world) )
       {
         frustum_cycle = true;
 
@@ -203,6 +206,10 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
           if(!this->ClearGridPoint(pt_index))
           {
             std::cout << "Failed to clear point." << std::endl;
+          }
+          else
+          {
+            updated_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
           }
         }
         else
@@ -228,11 +235,16 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
         {
           std::cout << "Failed to clear point." << std::endl;
         }
+        else
+        {
+          updated_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
+        }
         continue;
       }
     }
     // if here, we can add to costmap and PC2
     PopulateCostmapAndPointcloud(pt_index);
+    updated_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
   }
 }
 
--- a/src/spatio_temporal_voxel_layer.cpp
+++ b/src/spatio_temporal_voxel_layer.cpp
@@ -684,9 +684,10 @@ void SpatioTemporalVoxelLayer::UpdateROSCostmap( \
     }
   }
 
-  for (const auto& cell: cleared_cells)
+  std::unordered_set<volume_grid::occupany_cell>::iterator cell;
+  for (cell = cleared_cells.begin(); cell != cleared_cells.end(); ++cell)
   {
-    touch(cell.x, cell.y, min_x, min_y, max_x, max_y);
+    touch(cell->x, cell->y, min_x, min_y, max_x, max_y);
   }
 }
 
--- a/src/spatio_temporal_voxel_layer.cpp
+++ b/src/spatio_temporal_voxel_layer.cpp
@@ -665,24 +665,27 @@ void SpatioTemporalVoxelLayer::updateCosts( \
 void SpatioTemporalVoxelLayer::UpdateROSCostmap( \
                         double* min_x, double* min_y, \
                         double* max_x, double* max_y, \
-                        std::unordered_set<volume_grid::occupany_cell>& updated_cells)
+                        std::unordered_set<volume_grid::occupany_cell>& cleared_cells)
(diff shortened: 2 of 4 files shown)


src/spatio_temporal_voxel_grid.cpp
```diff
@@ -173,6 +175,7 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
   for (cit_grid; cit_grid.test(); ++cit_grid)
   {
     const openvdb::Coord pt_index(cit_grid.getCoord());
+    const openvdb::Vec3d pose_world = _grid->indexToWorld(pt_index);
```
SteveMacenski: Use the `this` version of this function. At some point I'll go back and make this function multithreaded and those functions were there for that purpose.

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -204,6 +207,10 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
           {
             std::cout << "Failed to clear point." << std::endl;
           }
+          else
+          {
+            updated_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
```
SteveMacenski: This could go before, not in, the `if` statement, the `if` is only for logging errors in multithreading. Same with L240 below.

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -228,11 +235,16 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
...
+        else
+        {
+          updated_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
+        }
         continue;
       }
     }
     // if here, we can add to costmap and PC2
     PopulateCostmapAndPointcloud(pt_index);
+    updated_cells.insert(occupany_cell(pose_world[0], pose_world[1]));
```
SteveMacenski: This represents the case where a point currenctly in the grid stays in the grid, this doesn't need to have the updated cells applied to it -- there's no change in this situation

SteveMacenski: The variable `updated_cells` really should be `cleared_cells` because its not including the newly marked cells, only the newly removed ones.

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -662,24 +662,28 @@ void SpatioTemporalVoxelLayer::updateCosts( \
...
+                        std::unordered_set<volume_grid::occupany_cell>& updated_cells)
 /*****************************************************************************/
 {
   // grabs map of occupied cells from grid and adds to costmap_
   Costmap2D::resetMaps();
 
-  std::unordered_map<volume_grid::occupany_cell, uint>::iterator it;
-  for (it = _voxel_grid->GetFlattenedCostmap()->begin();
-       it != _voxel_grid->GetFlattenedCostmap()->end(); ++it)
+  auto cost_map = _voxel_grid->GetFlattenedCostmap();
```
SteveMacenski: Revert this to iterators and specific calls, I don't understand why you changed this, it includes an entire copy

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -662,24 +662,28 @@ void SpatioTemporalVoxelLayer::updateCosts( \
...
+
+  for (const auto& cell: updated_cells)
```
SteveMacenski: This entirely breaks the clearing logic and is much slower. You're only checking the cleared cells, not the newly marked cells. What you should do here is revert this entire function to what it was and add a new `for` loop to `touch` the `cleared_cells` set. The marked cells set is implicitly covered by iterating through the entire flattened costmap. 

---

If you wanted to be very fancy, the "best" thing to do here would be to change this logic so that `costmap_` wasn't being cleared and re-populated each iteration. So you store the diffs of the clears and the diffs of the marks each iterations and you updated the `costmap_` by those. That would remove the entire flattened costmap iteration here but you'd need to store a `marked_cells` set too. I'm not advocating for you to do this, but that would be one way to do this.

SteveMacenski (changes requested):

src/spatio_temporal_voxel_grid.cpp
siferati: Ok. I copied the syntax from 

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/e077795c89a280ffffa83db61af540464e34a892/src/spatio_temporal_voxel_grid.cpp#L240-L245

Do you want me to replace it there as well?

src/spatio_temporal_voxel_grid.cpp
siferati: I did it like this because if the clearing failed, I don't see a reason to touch those cells - after all, they didn't change.

Do you mean that when multithreading `SpatioTemporalVoxelGrid::ClearGridPoint` might return false even though the clearing succeeded? If that's the case, then I agree that this shouldn't be inside the `else` statement.

src/spatio_temporal_voxel_grid.cpp
siferati: I agree that this represents the case where a point currently in the grid stays in the grid. However, because this is the only place where the flattened costmap is updated with obstacle information, it ends up representing all obstacles (i.e. marked cells) of the current cycle - which is why I store it in the `updated_cells` variable.

Please see my answer below for more context on this.

src/spatio_temporal_voxel_layer.cpp
siferati: Maybe I misunderstood how the marking logic is implemented, but it seems to me that obstacles are marked in the ROS costmap always 1 cycle after they are detected. *(I was going to create an issue about this, but wanted to focus on this PR before that).*

> The marked cells set is implicitly covered by iterating through the entire flattened costmap.

I agree with this. However, the flattened costmap is not up to date, but rather 1 cycle behind.

Obstacle information is only added to the flattened costmap in one place - inside the `SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud` function:

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/e077795c89a280ffffa83db61af540464e34a892/src/spatio_temporal_voxel_grid.cpp#L256-L266

And from what I understood, the call stack for that function is:

1. https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/e077795c89a280ffffa83db61af540464e34a892/src/spatio_temporal_voxel_grid.cpp#L235
2. https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/e077795c89a280ffffa83db61af540464e34a892/src/spatio_temporal_voxel_grid.cpp#L159
3. https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/e077795c89a280ffffa83db61af540464e34a892/src/spatio_temporal_voxel_layer.cpp#L722

Meaning obstacle information is added to the costmap before marking actually happens:

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/e077795c89a280ffffa83db61af540464e34a892/src/spatio_temporal_voxel_layer.cpp#L719-L743

When you call `SpatioTemporalVoxelGrid::Mark` (which in turn calls `SpatioTemporalVoxelGrid::operator()`), obstacle information is **stored in the voxel grid:**

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/e077795c89a280ffffa83db61af540464e34a892/src/spatio_temporal_voxel_grid.cpp#L317

However, it is **not stored in the flattened costmap** - this only happen during the next cycle, when `SpatioTemporalVoxelGrid::ClearFrustums` is called.

**TL;DR:** obstacle information is added to the costmap based on marking information from the previous cycle. In other words, the obstacle information of the costmap is always lagging 1 cycle behind.

---

That is why the variable `updated_cells` holds information about both clearing and marking. Information about cleared cells is stored when calling `SpatioTemporalVoxelGrid::ClearGridPoint`, and information about marked cells is stored when calling `SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud`. For both cases, this happens inside the function `SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap`.

Which is why I believe this PR doesn't actually break the current logic - it keeps it exactly the same. When you make the check `it->second >= _mark_threshold`, you are checking the flattened costmap that does not contain the most recent marking information - meaning it is the same information I stored in the `updated_cells` variable and thus we can avoid iterating the entire thing. 

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/e077795c89a280ffffa83db61af540464e34a892/src/spatio_temporal_voxel_layer.cpp#L672-L684

src/spatio_temporal_voxel_layer.cpp
siferati: I call the same function as you, `SpatioTemporalVoxelGrid::GetFlattenedCostmap`, which returns a pointer, not an entire copy. If anything, shouldn't my version be faster than yours, since I'm only calling that function once while you call it twice?

Then, I get the iterator I need by calling `unordered_map::find`, which runs in O(1) time.

Looping `updated_cells` is based on references, so I'm not making any copies either.

src/spatio_temporal_voxel_grid.cpp
SteveMacenski: Yes please

src/spatio_temporal_voxel_grid.cpp
SteveMacenski: > I did it like this because if the clearing failed, I don't see a reason to touch those cells - after all, they didn't change.

That's a totally fair assumption, but honestly I should remove that logging, it never triggers, it was more for development while there were situations where it was being called. Also, if we wanted to remove the point, we should also touch it for update in the costmap. These points are very likely grouped together so if 1 point is missed, its not like it will impact the bounds very much but we don't want to be in a situation where its problematic. 

Since this is just listing cells to touch for update -- and not actually updating their values directly -- we don't need to be as careful about it to have the `else` since it doesn't represent a direct update, just a request to _be_ looked at.

Tl;dr, please add before the `if()`

src/spatio_temporal_voxel_grid.cpp
SteveMacenski: You don't have to though, nothing has changed to need to be touched. Plus since this populates the flattened costmap, the `UpdateROSCostmap` function will take care of touching the cell for you https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/ros2/src/spatio_temporal_voxel_layer.cpp#L688

src/spatio_temporal_voxel_layer.cpp
SteveMacenski: O(1) doesn't mean it isn't slow, it just means its constant-slow, revert please. This is a huge amount of data you're trying to find and its going to ding performance. You'd rather loop through a vector of `M` values and do a thing than search `N` values, where`N` and `M` are similiarly sized due to processor caching.

src/spatio_temporal_voxel_layer.cpp
SteveMacenski: > TL;DR: obstacle information is added to the costmap based on marking information from the previous cycle. In other words, the obstacle information of the costmap is always lagging 1 cycle behind.

You are correct, I recall making that design choice way back when but looking at it now, it seems like the easy solution is just to call `PopulateCostmapAndPointcloud` (or similar) inside of `operator()` when it marks the point so that the `Mark` is adding to the cost map / grid points structures. Then you wouldn't have to change any of this. Not sure why I felt I needed to do that at the time, but it may have been before it was refactored into what we see today. 

It may be wise to make a new version of that for mark that doesn't convert into world coordinates (since you already have world coordinates) and only adds to `_grid_points` if it also inserts into costmap_ https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/ros2/src/spatio_temporal_voxel_grid.cpp#L230-L232. This is so that you don't have multiple `grid points` for the same voxelized cell (one from the voxel grid populated while clearing and another from marking) which would slow down performance a bunch if using a dense resolution sensor (you could have multiple returns in a single cell and end up with  `N` repeated instances artificially increasing the pointcloud2's size with redundant information)

src/spatio_temporal_voxel_layer.cpp
siferati: > you'd rather loop through a vector of M values and do a thing than search N values

There is no "searching" of any kind, you are not looping the map and checking every single element for its key.

`unordered_map` is an hash table. `unordered_map::find` simply computes the hash of the given key and returns the corresponding value. Hence why it's O(1) time complexity. Otherwise it would be O(N).

src/spatio_temporal_voxel_layer.cpp
siferati: Wouldn't the refactoring be better suited for a different PR? I think it would be cleaner to keep things separate: this PR fixes the issue of the incorrect clearing, and another PR refactors the marking logic.

src/spatio_temporal_voxel_layer.cpp
SteveMacenski: I don't think you understand. 

Doing O(1) N times can be slower than O(N) 1 time, you're equating orders of scale with the literal compute time. `find()` might be O(1) but its not a **fast** O(1), iterating through the entire container is nearly continuous memory so we can rely on caching the search can't. 

You can test this yourself using a chrono timer, I can almost guarantee this would be non-trivially slower. Looking at how algorithms scale as the container grows isn't the same as actual compute time.

src/spatio_temporal_voxel_layer.cpp
SteveMacenski: What refactor? You just call a function and its done, with no changes needed to the UpdateROSCostmap function. this seems to refactor far more than what I'm suggesting

src/spatio_temporal_voxel_grid.cpp
siferati: Fixed

src/spatio_temporal_voxel_grid.cpp
siferati: Fixed

src/spatio_temporal_voxel_grid.cpp
siferati: Fixed

src/spatio_temporal_voxel_layer.cpp
siferati: Ah, I see. Thanks for explaining!

src/spatio_temporal_voxel_layer.cpp
siferati: I meant more about splitting PRs by their intent. The intent of this PR was to fix the clearing issue, so IMO fixing the marking issue should be a different PR.

Anyway, fixed the marking too. Created a new overload for the function `SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud` that receives a world pose.

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -716,10 +723,12 @@ void SpatioTemporalVoxelLayer::updateBounds( \
   ObservationsResetAfterReading();
   current_ = current;
 
+  std::unordered_set<volume_grid::occupany_cell> cleared_cells;
```
SteveMacenski, line 727: Why an `unordered_set`? Why not a vector for instance? Was this experimentally shown to be faster?

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -681,6 +683,11 @@ void SpatioTemporalVoxelLayer::UpdateROSCostmap(double* min_x, double* min_y, \
       touch(it->first.x, it->first.y, min_x, min_y, max_x, max_y);
     }
   }
+
+  for (const auto& cell: cleared_cells)
```
SteveMacenski: Just for styling, can we iterate through this using iterators? It makes life easy for trying different container types for run-time speeds

SteveMacenski (commented):
A couple of questions, but overall LGTM now. Next steps after resolving these questions is to cherry pick this commit to noetic / ros2 / foxy

src/spatio_temporal_voxel_layer.cpp
siferati: Fixed

src/spatio_temporal_voxel_layer.cpp
siferati: I chose `unordered_set` to avoid duplicates. When clearing the world points, we don't care about the Z coordinate. If we clear world point (1, 2, 3) and (1, 2, 4), they're both gonna be in the same costmap cell, so there is no need to store and iterate through it twice.

SteveMacenski: OK, that  seems reasonable. I'd ask though if you'd test both and use which ever is practically faster. The STL Vector implementation is very cache friendly, so even areas where you wouldn't think it would be faster to use (because of duplications for instance), it ends up being faster due to continuous memory buffers (I ran into that alot while developing this: what you "think" should be faster based on theory and what is actually faster are not always intuitive). I don't know if it would be faster or not, so please test and let me know. 

You can create an update timer pretty easily with chrono 
```
StartTime = std::chrono::system_clock::now();
... do work
endTime = std::chrono::system_clock::now();
std::cout << "time took " std::chrono::duration_cast<std::chrono::milliseconds>(endTime - StartTime).count() << " ms."<< std::endl;
```

siferati:
I think there's something wrong with the way I add the new marked points to `_grid_points`. I'm seeing a lot more points in the point cloud in RViz. 

If both the robot and the obstacle are static, the point cloud should be almost the same every cycle.

However, with this PR the point cloud is **a lot** denser than before, meaning the points I add when marking have different coordinates than the ones that are added on clearing.

SteveMacenski:
Not sure, it looks OK to me, you should only see the new points from the measurement sensor that weren't in the grid already.

Are they overlapping each other? The only thing I can think of is that since you input the global coordinates directly from the sensor, that these points weren't already in the `_cost_map` because they're not binned into the voxels. You could try binning them before calling `PopulateCostmapAndPointcloud`. A simple way to prototype would be to convert to index and back to global. The other points are coming out of the voxel grid so they'll have been binned by the voxelization transform.

siferati:
>  because they're not binned into the voxels

Yup, that was the problem. I compared the size of `_grid_points` and `_cost_map` when the only visible obstacle was the wall on the right (see video in OP). There's a very big difference between their sizes:

* binned `_grid_points->size()`: ~4000
* binned `_cost_map->size()`: ~100
* not binned `_grid_points->size()`: ~8000
* not binned `_cost_map->size()`: ~4000

Since it seems like we can't input the global coordinates directly, I removed my overload of the `SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud` function and edited the original one to handle both clearing and marking.

I think the only problem left is the one I mentioned in the OP about the sides of the costmap not being properly cleaned.

src/spatio_temporal_voxel_layer.cpp
siferati: You're right, `vector` is **a lot** faster than `unordered_set` when iterating through its contents, even though its size is a lot bigger due to duplicated elements. However, `vector` is only faster if we don't do anything while iterating. Since we need to touch every element, then `unordered_set` becomes the faster alternative since it has fewer elements - meaning less calls to `touch`.

I recorded two small clips where you can see the time it takes for both cases, when decay time is set to 0 (i.e. instant clearing):

* [looping only](https://streamable.com/3mi5zd)
* [looping and touching every element](https://streamable.com/6ysvm6)

SteveMacenski: What was the looping testing? Just the loop without `touch` (so empty `for`?) 

Keep in mind its not just access for iteration, its also about insertion time since vector insertion would probably be faster due to no hash table lookups for existing keys. You should try to measure the full update and take some metrics based on that since the insertions are pretty built into the clearing function. I agree its looking good for the `unordered_set` but just want to validate. I did alot of testing on datastructures to make STVL real-time so I want to ensure this doesn't have an undo impact on the overall performance.

So tl;dr
- Measure this update cycle metric
- Measure this with vector update cycle metric
- Measure the baseline before-you-changed-anything metric

So we can answer the questions:
- Is vector or unordered_set the right answer?
- What impact does this have on the update time compared to what existed before?

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -237,19 +237,20 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
...
+void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(                   \
+                                                   const openvdb::Coord& pt,  \
+                                                   const bool& pub_every_voxel)
 /*****************************************************************************/
 {
   // add pt to the pointcloud and costmap
   openvdb::Vec3d pose_world = this->IndexToWorld(pt);
 
-  if (_pub_voxels)
+  if (_pub_voxels && pub_every_voxel)
```
SteveMacenski, line 253: This doesn't make sense - you'd now be inserting it twice if its new. Once here and once below in the `else` statement. Please look over the functions after you prototype something to make sure it makes sense

SteveMacenski (changes requested):

SteveMacenski:
I agree that since we have to convert from global->index->global to artificially bin it, it makes sense to use the same function, but your choice of implementation is flawed

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -232,19 +237,20 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
...
     // if here, we can add to costmap and PC2
-    PopulateCostmapAndPointcloud(pt_index);
+    PopulateCostmapAndPointcloud(pt_index, true);
   }
 }
 
 /*****************************************************************************/
-void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(const \
-                                                            openvdb::Coord& pt)
+void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(                   \
```
SteveMacenski: I think this would make more semantic sense if:

- under `if (cell != _cost_map->end())` we published if `_pub_voxels && pub_every_voxel` since its not a new insert, so only publish if meant to publish and that every voxel param is true.
- then in the `else` you can change to as `_pub_voxels`, no need to even check the param because its a pure insert either way.

SteveMacenski (changes requested):

SteveMacenski:
> I think the only problem left is the one I mentioned in the OP about the sides of the costmap not being properly cleaned.

I don't recall that, what's that issue

siferati:
> This mostly fixes the issue, however there is still an edge case I wasn't able to find a solution for. When the cleared obstacles are on the side, then the local costmap remains uncleared. You can see this happening in [this](https://streamable.com/lim2bh) video: although the obstacle in the front was correctly cleared, the obstacle in the left side remains.

SteveMacenski:
Were those points outside of the bounds of the local costmap? If so, they shouldn't be marking because they shouldn't be `touched` to be able to expand the bounds outside of that window range. Is this new to this branch? If so, that means you're `touch`ing things that shouldn't be touched expanding the bounds.

src/spatio_temporal_voxel_layer.cpp
siferati: I averaged 450 update cycles with the following scenario (decay time set to 0):

image


The start time was set right before defining the `cleared_cells` variable: https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/651fcd7762fa86f7defa6e20b18ac21933be37c3/src/spatio_temporal_voxel_layer.cpp#L727

and end time at the very bottom of the function, right before `return`.

If I set the start time at the top of the function the execution times were **very** inconsistent, I'm guessing due to the calls to `GetObservations` methods. I don't think my changes should affect the performance before we clear anyway, so I set the start time after the observations are obtained, in order to get more consistent times.

Results (units in microseconds):

* **baseline**
  * mean: 2983
  * std: 430
* **unordered_set**
  * mean: 3761
  * std: 921
* **vector**
  * mean: 3512
  * std: 836

So `unordered_set` and `vector` have basically the same performance, which was ~0.5ms worse than the baseline.

src/spatio_temporal_voxel_grid.cpp
siferati: Fixed

siferati:
> Were those points outside of the bounds of the local costmap?

Nop

> Is this new to this branch?

The upstream branch didn't clear the obstacle at all. This PR fixes some of that.

I did some more testing and the behavior is very strange. It seems like the area that is not cleared is based on the fixed map frame coordinates. I initially thought the obstacles on the side of the robot weren't getting cleared, but that's not the case - it's the obstacles that are past some coordinate in the map frame. In the video I have in the OP, you can see obstacles in the front and right are properly cleared, while the obstacle on the left remains.

If I move the robot towards the left, then all obstacles will stop being cleared. In the video, the robot is at position (58, 28), and all obstacles except that small bit on the left are properly cleared. However, if I move the robot a bit more towards the left, for example (53, 28), then the costmap is no longer cleared and faces the same issue I had with the upstream branch, #199 .

siferati:
> Were those points outside of the bounds of the local costmap?

Nop.

> Is this new to this branch?

The upstream branch didn't clear the obstacles at all. This PR at least clears them under certain conditions, although it seems like the main problem is still not solved.

After some more testing, I think the problem is that in the video I posted in the OP there were no walls on the left side. Take a look at [this](https://streamable.com/8tfl8v) video, where the robot is in the middle of the map with no walls near it. The obstacles stay permanently in the costmap until a different obstacle is detected in the same direction behind it.

SteveMacenski:
How does this PR not solve that issue? If your caching the clears, then it should be part of the next update bounds update to get queried and cleared out

siferati:
> How does this PR not solve that issue?

I don't know...! From what I can tell, there are only two places where the voxels are cleared, and I'm caching the cells in both cases:

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/651fcd7762fa86f7defa6e20b18ac21933be37c3/src/spatio_temporal_voxel_grid.cpp#L206-L207
https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/651fcd7762fa86f7defa6e20b18ac21933be37c3/src/spatio_temporal_voxel_grid.cpp#L231-L232

The current state of affairs is:

* upstream: obstacles are never removed from the costmap, issue #199 
* this PR: obstacles are removed from the costmap only if there is another obstacle behind it (even if that other obstacle is outside the costmap bounds)
* In both cases, putting a static layer under the STVL layer fixes the problem

SteveMacenski:
Make any progress on this?

siferati:
Hi, unfortunately I wasn't able to find the cause of the issue. Since touching the cleared cells doesn't seem to fix it, then maybe the underlying problem is something different from your initial guess?

If you have any ideas I'd be happy to try them out! Otherwise I can also provide a minimum working example if you'd like.

SteveMacenski:
@siferati is it as simple as https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/651fcd7762fa86f7defa6e20b18ac21933be37c3/src/spatio_temporal_voxel_layer.cpp#L690 this is not also setting the cell as cleared? https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/651fcd7762fa86f7defa6e20b18ac21933be37c3/src/spatio_temporal_voxel_layer.cpp#L682

Keep in mind there could be multiple cells in that column that could still be marked so it shouldn't be always clearing `costmap_`, but if there's no more items marked in that column it should be, e.g.

```
  std::unordered_set<volume_grid::occupany_cell>::iterator cell;
  for (cell = cleared_cells.begin(); cell != cleared_cells.end(); ++cell)
  {
    if (column contains < _mark_threshold) {
      costmap_[getIndex(map_x, map_y)] = costmap_2d::FREE_SPACE; // You still need to retrieve map_x and map_y or the index
      touch(cell->x, cell->y, min_x, min_y, max_x, max_y);
    }

  }
}
```

siferati:
> Keep in mind there could be multiple cells in that column that could still be marked so it shouldn't be always clearing costmap_, but if there's no more items marked in that column it should be

The problem remains even if I do that. But while testing your idea I found something very interesting!

a) If I mark cleared cells as free space **before** lethal obstacles are marked, the problem remains:

```cpp
void SpatioTemporalVoxelLayer::UpdateROSCostmap( \
                        double* min_x, double* min_y, \
                        double* max_x, double* max_y, \
                        std::unordered_set<volume_grid::occupany_cell>& cleared_cells)
{
  // grabs map of occupied cells from grid and adds to costmap_
  Costmap2D::resetMaps();

  std::unordered_set<volume_grid::occupany_cell>::iterator cell;
  for (cell = cleared_cells.begin(); cell != cleared_cells.end(); ++cell)
  {
    uint map_x, map_y;
    if (worldToMap(cell->x, cell->y, map_x, map_y))
    {
      costmap_[getIndex(map_x, map_y)] = costmap_2d::FREE_SPACE;
      touch(cell->x, cell->y, min_x, min_y, max_x, max_y);
    }
  }

  std::unordered_map<volume_grid::occupany_cell, uint>::iterator it;
  for (it = _voxel_grid->GetFlattenedCostmap()->begin();
       it != _voxel_grid->GetFlattenedCostmap()->end(); ++it)
  {
    uint map_x, map_y;
    if ( it->second >= _mark_threshold && \
         worldToMap(it->first.x, it->first.y, map_x, map_y))
    {
      costmap_[getIndex(map_x, map_y)] = costmap_2d::LETHAL_OBSTACLE;
      touch(it->first.x, it->first.y, min_x, min_y, max_x, max_y);
    }
  }
}
```

b) However, if I mark cleared cells as free space **after** lethal obstacles are marked, the problem is fixed:

```cpp
void SpatioTemporalVoxelLayer::UpdateROSCostmap( \
                        double* min_x, double* min_y, \
                        double* max_x, double* max_y, \
                        std::unordered_set<volume_grid::occupany_cell>& cleared_cells)
{
  // grabs map of occupied cells from grid and adds to costmap_
  Costmap2D::resetMaps();

  std::unordered_map<volume_grid::occupany_cell, uint>::iterator it;
  for (it = _voxel_grid->GetFlattenedCostmap()->begin();
       it != _voxel_grid->GetFlattenedCostmap()->end(); ++it)
  {
    uint map_x, map_y;
    if ( it->second >= _mark_threshold && \
         worldToMap(it->first.x, it->first.y, map_x, map_y))
    {
      costmap_[getIndex(map_x, map_y)] = costmap_2d::LETHAL_OBSTACLE;
      touch(it->first.x, it->first.y, min_x, min_y, max_x, max_y);
    }
  }

  std::unordered_set<volume_grid::occupany_cell>::iterator cell;
  for (cell = cleared_cells.begin(); cell != cleared_cells.end(); ++cell)
  {
    uint map_x, map_y;
    if (worldToMap(cell->x, cell->y, map_x, map_y))
    {
      costmap_[getIndex(map_x, map_y)] = costmap_2d::FREE_SPACE;
      touch(cell->x, cell->y, min_x, min_y, max_x, max_y);
    }
  }
}
```

Since b) works but both a) and your suggestion don't, it seems to me that the problem is that the voxel flattened costmap contains cells that should have been cleared but weren't. Or rather, since
https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/e077795c89a280ffffa83db61af540464e34a892/src/spatio_temporal_voxel_grid.cpp#L114
is called every update, it seems more logical to say that there is obstacle information being added to the costmap about obstacles that no longer exist.

SteveMacenski:
I'll need to think about that a little.

```
    if (worldToMap(cell->x, cell->y, map_x, map_y))
    {
      costmap_[getIndex(map_x, map_y)] = costmap_2d::FREE_SPACE;
      touch(cell->x, cell->y, min_x, min_y, max_x, max_y);
    }
```

This isn't quite right I don't think. You didn't check if the cleared voxel clears enough of the column of the voxels to make it as free now. If you had a column of 10 voxels and you cleared one, you shouldn't mark the costmap now as clear, its still occupied. You need to check for the number of cells still marked in the column and if this clearing operation now brings the number of voxels marked in this column of the costmap to `0` or below the mark threshold. 

Though, I have to admit, I would need to sketch it out, but the logic of placing the clearing after marking **might** implicitly solve that for you since we only mark as LETHAL when the mark threshold is already met (and is cleared every iteration). In fact, since we're clearing the costmap every single iteration (`resetMaps`) and the  `GetFlattenedCostmap()` is also repopulated every single iteration, I'm not wondering why any of this `cleared_cells` stuff is needed. If we clear everything to `FREE` and we only set costmap cells to `LETHAL` if  `GetFlattenedCostmap()` points are higher than `_mark_threshold`, there should never be a situation where things are over marked. 

Where is the state being tracked between iterations incorrectly needing any of this?  I don't think that block of code I suggested above should actually be there, if that's working, that means something is marking points that shouldn't have been marked at all. 

- `_cost_map` is cleared at the start of `clearFrustums` thusly emptying the flattened costmap every iteration
- `_cost_map` is updated in clearing frustums only when the point its iterating over remains in the VDB grid. If we `ClearPoint`, it never gets to the `Populate...` function to insert itself into `_cost_map`. We +1 count on the costmap cell. 
- `_cost_map` is then updated when a new point is marked with new data with a +1 count from an addition you made in this PR
- Then in `updateBounds`, we iterate over `_cost_map` to fill in an empty `costmap_` object with `LETHAL` only when the count in an individual cell is high enough

I don't see where we went wrong that there would be anything marked in the `costmap_` that wasn't in the VDB grid.

The only slight hiccup I see is that when you added `PopulateCostmapAndPointcloud()` to the Marking functions, there's no check on if the 3D (x,y,z) voxelized point that is being viewed by a sensor in marking before inserting it into the `_cost_map` and incrementing the column count +1. What that could do is inflate the count, since you're +1-ing a point that was already +1-ed in the grid already from the clearing updates. Those counts are supposed to be for the number of viewed **voxels**, not the number of total column hits. 

I'm trying to follow that train of logic and I don't see how that would result in over marking if the obstacle simply vanished (because there would be no marking at that cell to cause it), but maybe that's what's actually causing this?

siferati:
> You didn't check if the cleared voxel clears enough of the column of the voxels to make it as free now

> The problem remains even if I do that
> both a) and your suggestion don't [work]

I did it according to your suggestion. I just didn't post the code snippet since you had already posted the pseudo-code above. And since marking **every** cleared cell as free space **before** lethal obstacles are marked doesn't fix the problem, it makes sense that your suggestion also wouldn't work. Marking every cleared cell as free space and then overwriting the ones that are above a threshold with lethal obstacles would result in the same costmap as first marking the cells above the threshold as lethal obstacles and then marking as free space the cells below that threshold.

> wondering why any of this cleared_cells stuff is needed

Because we need to touch the cleared cells to update the bounds. Quoting your previous comment: *the internal voxel grid is empty on clearing, then nothing new marks it because the obstacle instantly disappeared, and then when it comes time to update the bounds, there's nothing to update them with (and there's no static layer to set the reasonable update bounds).*

> The only slight hiccup I see is that when you added PopulateCostmapAndPointcloud() to the Marking functions, there's no check on if the 3D (x,y,z) voxelized point that is being viewed by a sensor in marking before inserting it into the _cost_map and incrementing the column count +1. What that could do is inflate the count, since you're +1-ing a point that was already +1-ed in the grid already from the clearing updates.

The only way I can think of to solve this is to keep track of the points that we marked in the clearing update in some container, and then when marking new points check for inclusion in that container: if it exists, then we skip the first +1. But this would likely slow down the process quite a bit, since for every marking observation we would have to check if it exists in that container. I'm not sure if it's worth doing this in order to avoid the column count sometimes being off by 1.

---

### I finally found the problem!

> If we ClearPoint, it never gets to the Populate... function to insert itself into _cost_map.

When the point is cleared here

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/e077795c89a280ffffa83db61af540464e34a892/src/spatio_temporal_voxel_grid.cpp#L227-L232

we `continue` to the next point. But when we clear the point here

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/e077795c89a280ffffa83db61af540464e34a892/src/spatio_temporal_voxel_grid.cpp#L203-L206

we do nothing. So despite clearing this point, we will still populate it in 

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/e077795c89a280ffffa83db61af540464e34a892/src/spatio_temporal_voxel_grid.cpp#L234-L235

I added a flag to avoid calling `PopulateCostmapAndPointcloud` with points we cleared and that solved the problem.

SteveMacenski:
Rather than `cleared_point`, could you not just `break;` on L212 after we clear it so that we never enter that part of the loop like we do with Marking?

That also makes me wonder if we also break on L238 after the other clearing operation, then we never hit the `PopulateCostmapAndPointcloud` call, and thusly cleared points wouldn't entire that pointcloud. I think that could be the bug requiring `cleared_cells` (we might be able to remove this now?)

I think the fact that I didn't add `break;`s after the clears and only the marking operation in ClearFustums is the cause of the issue. If we add that breaks, I think everything else would fall back in line, no? It looks like L238 had a continue before your PR changes, but I think maybe that continue should actually be a break (but either way, in that situation it didn't hit populate; but the other clear did)

siferati:
> Rather than cleared_point, could you not just break; on L212 after we clear it so that we never enter that part of the loop like we do with Marking?

If you `break` there, then you're breaking the inner loop (frustums), not the outer loop (grid points). That's what's happening with marking, you break out of the inner loop -- otherwise you would never reach the `PopulateCostmapAndPointcloud` call. That's why the `cleared_point` flag is necessary.

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/e077795c89a280ffffa83db61af540464e34a892/src/spatio_temporal_voxel_grid.cpp#L171-L185

> That also makes me wonder if we also break on L238 after the other clearing operation, then we never hit the PopulateCostmapAndPointcloud call, and thusly cleared points wouldn't entire that pointcloud.

> It looks like L238 had a continue before your PR changes, but I think maybe that continue should actually be a break

If you break on L238, then you're breaking out of the outer loop (grid points), meaning you will return from the function and not evaluate the remaining points. That's why there was a `continue` statement there previously -- to skip the call to `PopulateCostmapAndPointcloud` but keep iterating through the remaining points.

I simply replaced the `continue` statement with the `cleared_point` flag for clarity's sake, so that it would have the same syntax as the 1st clear above, and so that I would only have to write the `cleared_cells.insert` statement once -- thus avoiding duplicated code.

--- 

>> The only slight hiccup I see is that when you added PopulateCostmapAndPointcloud() to the Marking functions, there's no check on if the 3D (x,y,z) voxelized point that is being viewed by a sensor in marking before inserting it into the _cost_map and incrementing the column count +1. What that could do is inflate the count, since you're +1-ing a point that was already +1-ed in the grid already from the clearing updates. Those counts are supposed to be for the number of viewed voxels, not the number of total column hits.

> The only way I can think of to solve this is to keep track of the points that we marked in the clearing update in some container, and then when marking new points check for inclusion in that container: if it exists, then we skip the first +1. But this would likely slow down the process quite a bit, since for every marking observation we would have to check if it exists in that container. I'm not sure if it's worth doing this in order to avoid the column count sometimes being off by 1.

I thought a bit more about this and realized that it is actually not just off by 1, but off by **double**. If we have some object that is made up of, for example, 10 voxels in a vertical line (so they all share the same (x, y) coords), then what's gonna happen is that in the clearing update we're gonna update the corresponding costmap cell 10 times, resulting in the cell having a value of 10, and then when marking we're gonna update it another 10 times -- resulting in that cell having a column value of 20!

Given the above, then I think it is better to either fix it or revert the marking logic so that it stays 1 update behind (i.e. the current upstream behaviour, before this PR).

What do you think?

SteveMacenski:
> If you break there, then you're breaking the inner loop (frustums), not the outer loop (grid points).

Yes, that would be the point, then you move onto the next point, which would then have the opportunity to be populated. But that would `break` out of the current `point` for all future frustums (since if you just **cleared** it, it doesn't matter what any other frustum does in those contexts). 

I think, unless I'm missing something, that flag isn't necessary. 

I agree with you on L238, that should be a `continue` not a `break`, good catch. The `continue` is what was there before that it looks like your PR removed. 

I do really think though from the looking I did this morning is if we re-add the `continue` on L238 and add that `break` I suggested in the frustum clearing, then we no longer require `clearing_cells` at all. We needed that before because we weren't breaking out of the loop, so we were adding cells that were just cleared into the `Populate...` function. If we break, then the `Populate...` never gets cleared cells, and thusly there's no need to track the specifically cleared frustum cells anymore.

From the behavior you reported too, that logic tracks for me. The reason it didn't clear is because it was still in the frustum view, and when we cleared, we didnt `break;` so we ended up adding that cell to `Populate...` which then in turn was added to the costmap. 


>  resulting in the cell having a value of 10, and then when marking we're gonna update it another 10 times -- resulting in that cell having a column value of 20!

Exactly, that's why I mentioned that since you added the `Populate...` function to marking, we need to check if that cell was already in the grid and only increment the count if its a new cell. That may be possible to do without adding too much overhead. Either way, even if we keep the updates to 1 cycle off, we should still fix some of these other concerns you've brought up with not having a static map underneath it. But I would tend to agree we should fix 1 issue at a time. Resolve the main issue first and then the 1 cycle off issue later.

siferati:
> But that would break out of the current point for all future frustums (since if you just cleared it, it doesn't matter what any other frustum does in those contexts).

Ok, I agree. Adding the `break` statement for this reason makes sense.

> Yes, that would be the point, then you move onto the next point, which would then have the opportunity to be populated.

> I think, unless I'm missing something, that flag isn't necessary.

By inserting the `break` statement there, you're exiting the inner frustum loop, but you're still inside the outer loop. So the next statement to be evaluated is this:

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/e077795c89a280ffffa83db61af540464e34a892/src/spatio_temporal_voxel_grid.cpp#L222-L233

Since `frustum_cycle` was set to `true` above, then you do **not** enter that if statement. The next statement evaluated is 

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/e077795c89a280ffffa83db61af540464e34a892/src/spatio_temporal_voxel_grid.cpp#L234-L235

So you will still populate the costmap with the cleared point. In order to avoid this, you would need to `break` the inner loop and immediately `continue` the outer loop. This is not possible without using some sort of flag.

> But I would tend to agree we should fix 1 issue at a time. Resolve the main issue first and then the 1 cycle off issue later.

Agreed.

SteveMacenski:
Ah yes, we break out of the inner loop, it doesn't just reset to the outer, it still has that code left. OK. A bool `point_cleared` then could be added if we clear it there and then we can add a 

```
if (cleared_point) {
  continue;
}
```
 
after the `for` loop but before the `if(!frustum_cycle) {...}` statement. I think _now_ that might fix our issues, yes? :wink:

siferati:
But that's exactly the same behavior of what I have already done... The only difference is that with the way you're suggesting I would have to write `cleared_cells.insert` twice. The way I did it, I only write it once and the code is easier to read.

Edit: only thing missing is to add the `break` statement we discussed above -- I'll do that tomorrow

SteveMacenski:
Well I think we've found from this method, we don't require `cleared_cells` at all anymore, yeah? If we don't include them in `Populate...` then we're not artificially having more than we should to need to clear them out?

siferati:
> we don't require cleared_cells at all anymore

We've already discussed this a few comments above:

>> wondering why any of this cleared_cells stuff is needed

> Because we need to touch the cleared cells to update the bounds. Quoting your previous comment: *the internal voxel grid is empty on clearing, then nothing new marks it because the obstacle instantly disappeared, and then when it comes time to update the bounds, there's nothing to update them with (and there's no static layer to set the reasonable update bounds).*

There are 3 different problems at play here:

1. Cleared cells are not touched, so the corresponding area of the costmap isn't updated. The `cleared_cells` container solves this.
2. Cleared cells are still marked in the cosmap as obstacles. This `cleared_point` flag solves this.
3. Marking is 1 update cycle behind. I reverted this PR to not address this, as it is more complicated than initially assumed and should be handled in its own PR.

`cleared_cells` and `cleared_point` are not interchangeable and both solve different problems.

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -200,10 +204,12 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
         if (time_until_decay < 0.)
         {
           // expired by acceleration
+          cleared_point = true;
           if(!this->ClearGridPoint(pt_index))
           {
             std::cout << "Failed to clear point." << std::endl;
           }
+          break;
```
SteveMacenski, line 212: remove this break, I believe I was wrong here. What should be happening is that for each clearing sensor it should be able to accelerate the points to handle multiple-overlapping frustums to allow more significant clearing. 

This is a really subtle function that I haven't been developing on for a few years now, apologies for some of the flip-flopping.

SteveMacenski (commented):
OK, I think the rest of this tracks for me after that change. Have you tested this that it resolves the issues and doesn't break anything?

src/spatio_temporal_voxel_grid.cpp
siferati: That `break` only happens when the point is cleared from the grid due to its decay time being less than 0 (it's inside the `if (time_until_decay < 0.)` block). In this case, there is no need to check other frustums -- even if another overlapping frustum would further accelerate the decay, the point was already cleared from the grid. I think your initial reasoning for adding this `break` statement is correct.

SteveMacenski: :facepalm: clearly I haven't looked at this in depth recently.

siferati:
> Have you tested this that it resolves the issues and doesn't break anything?

This PR fixes #199. I didn't find any problems when testing.

SteveMacenski:
OK here's what I recommend for next steps:
- Submit this patch to Noetic, Foxy, and ROS2 branches so that the newest distributions can get these patches as well
- I'm going to reach out to some users of STVL in production to test this patch to make sure they don't run into any issues. I'd test myself, but I don't use this anymore in a product (I've since left that company and I'm not working with product robots anymore) so I want to make sure that others that this more directly impacts have a say
- And assuming all is well, we can merge! 

---

@doisyg can you test this either in ROS1 or ROS2 (once a PR for ROS2 is made)? Just to sanity check this doesn't break anything for you?

siferati:
> Submit this patch to Noetic, Foxy, and ROS2 branches

Done. But I don't have access to a test environment for those distributions, so I wasn't able to test if things *actually* worked -- I'll leave that to someone else.

doisyg:
I may have some bandwith to do a RO2 sanity check on our hw platform (which uses 2 3D cameras in STVL) at the end of the week. Please ping me back then if I don't

SteveMacenski:
@doisyg friendly ping :-)

Patrick-Lascombe:
I tested the PR on our ROS2 hardware, it solves the issue without impacting the behaviour of the STVL

SteveMacenski:
Sweet, merging!

SteveMacenski (approved):

PR #203, opened 2021-05-15
Applied Melodic fix #200 to ROS2

siferati:
See #200 for further details.


SteveMacenski:
Note for self: cherry pick this to Galactic

SteveMacenski (approved):

PR #205, opened 2021-06-11
Replace ROS WallTime with normal ROS Time

Flova:
This fixes issues when the code runs in a slow simulation. Currently, obstacles are removed according to the wall clock which leads to a shorter lifespan in the simulation time.

### DIFF
--- a/src/spatio_temporal_voxel_grid.cpp
+++ b/src/spatio_temporal_voxel_grid.cpp
@@ -168,7 +168,7 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
 /*****************************************************************************/
 {
   // sample time once for all clearing readings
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):

SteveMacenski:
Unfortunately, that would break existing behavior for users, so this cannot be merged into already released distributions. I think its a fine thing to do, but it cannot be merged back into already released and maintained distributions for fear that it would severely impact users in the field.

timonegk:
Maybe it could be configurable? So that the default is walltime but it would be possible to set a parameter to use ros time.

SteveMacenski:
After further review, I think this is a good idea - merging. Sorry for the delay

SteveMacenski (approved):

PR #213, opened 2021-10-01
Fix voxel visualization

doisyg:
Voxel visualization currently outputs a pointcloud where the x and y coordinates of each point are the coordinates of the corner of the voxel. This can be confusing while debugging in rviz as illustrated below: 
- costmap lethal pixels are in pink
- input stvl sensor pointcloud is the small blue spheres (one is green due to transparency)
- voxel visualization is in green and displayed as 0.05 boxes with alpha 0.4
- both the costmap and stvl have a resolution of 0.05
voxels
One should naively expect to have the green boxes and the costmap pixels aligned.

This PR proposes to change the behavior by centering the points of the voxel visualization pointcloud on the voxels they are suppose to represent.

### DIFF
--- a/src/spatio_temporal_voxel_grid.cpp
+++ b/src/spatio_temporal_voxel_grid.cpp
@@ -355,13 +355,14 @@ void SpatioTemporalVoxelGrid::GetOccupancyPointCloud(
   sensor_msgs::PointCloud2Iterator<float> iter_y(*pc2, "y");
   sensor_msgs::PointCloud2Iterator<float> iter_z(*pc2, "z");
 
+  double center_offset = _voxel_size/2;
   for (std::vector<geometry_msgs::msg::Point32>::iterator it =
     _grid_points->begin();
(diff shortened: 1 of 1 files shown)


src/spatio_temporal_voxel_grid.cpp
```diff
@@ -355,13 +355,14 @@ void SpatioTemporalVoxelGrid::GetOccupancyPointCloud(
   sensor_msgs::PointCloud2Iterator<float> iter_y(*pc2, "y");
   sensor_msgs::PointCloud2Iterator<float> iter_z(*pc2, "z");
 
+  double center_offset = _voxel_size/2;
```
SteveMacenski:
```suggestion
  const double & center_offset = _voxel_size / 2.0;
```

SteveMacenski (changes requested):
Minor change and A-OK

SteveMacenski (approved):

Issue #220, opened 2022-01-17
ClearCostmapExceptRegion not clearing voxels


lowellausen:
Hello!

I believe I have found some bug/unexpected behavior from the spatio temporal voxel layer (actually highly related to nav2 costmap2d as well, was unsure where to post this first). 

It seems that we have a few services for recovery behaviors that clear the costmaps in order to remove artifacts, one of them is for example `/global_costmap/clear_entirely_global_costmap` and the other is `/global_costmap/clear_except_global_costmap`. 

When I call the `ClearEntireCostmap` I get the expected behavior that all the not currently visible voxels are cleared along with their costmaps, but when I call the `ClearCostmapExceptRegion`  with whatever value for `reset_distance` parameter (have tried for example 2.0, 1.0, 0.5, 0.0) none of the expected voxels/costmaps are cleared.

Looking into the code I've realized that `ClearEntireCostmap` will call a reset function that is actually implemented in stvl:

The reset function for one layer (stvl in this case)

Is implemented here:  https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/396b25b806b22d9c256aa2000691f5dd171b4571/src/spatio_temporal_voxel_layer.cpp#L545-L557

Called here: https://github.com/ros-planning/navigation2/blob/8a8b4d2ff97b36ae9931be376fe3ddd701d19445/nav2_costmap_2d/src/costmap_2d_ros.cpp#L518-L531

And defined as virtual here: https://github.com/ros-planning/navigation2/blob/b201d058a7b15f1786c57afe69d079ac5d502018/nav2_costmap_2d/include/nav2_costmap_2d/layer.hpp#L84-L87

Which so far makes a lot of sense, as this works.

But when looking at how the `ClearCostmapExceptRegion` clears the costmap it is in quite a different way. It doesn't use anything similar to a reset function, instead it uses this `resetMapToValue`

That is implemented directly in costmap_2d: https://github.com/ros-planning/navigation2/blob/8a8b4d2ff97b36ae9931be376fe3ddd701d19445/nav2_costmap_2d/src/costmap_2d.cpp#L102-L110

And is defined here not as virtual like the reset one: https://github.com/ros-planning/navigation2/blob/8a8b4d2ff97b36ae9931be376fe3ddd701d19445/nav2_costmap_2d/include/nav2_costmap_2d/costmap_2d.hpp#L311-L312

[...]

PR #223, opened 2022-01-31
Adding ClearGridExceptRegion service (#220)

ghost:
Hello!

As we discussed on #220, I've created a service for clearing the grid except for a safety square region around the robot. The region is defined with the parameter reset_distance that is the side of the square, as its also defined in the ClearExceptRegion on navigation2 side.

I won't say I'm 100% happy with this implementation, as I had to do something ugly to get access to robot position when doing the clearing because we don't have access to it at all times (hence when we call the service (I think)). But this is working for us!

So we've decided to open this PR anyways, see if it's of your interest to merge it, or if there are better ideas for the implementation before the merge.

### DIFF
--- a/src/spatio_temporal_voxel_grid.cpp
+++ b/src/spatio_temporal_voxel_grid.cpp
@@ -451,4 +451,51 @@ bool SpatioTemporalVoxelGrid::SaveGrid(
   return false;  // best offense is a good defense
 }
 
+/*****************************************************************************/
+void SpatioTemporalVoxelGrid::ClearGridExceptRegion(
+  double robot_x, double robot_y, double reset_distance,
+  std::unordered_set<occupany_cell> & cleared_cells)
+/*****************************************************************************/
+{
+  boost::unique_lock<boost::mutex> lock(_grid_lock);
+
+  if (this->IsGridEmpty()) {
+    return;
+  }
+
+  _grid_points->clear();
+  _cost_map->clear();
+
+  double start_point_x = robot_x - reset_distance / 2;
+  double start_point_y = robot_y - reset_distance / 2;
+  double end_point_x = start_point_x + reset_distance;
+  double end_point_y = start_point_y + reset_distance;
+
+  openvdb::DoubleGrid::ValueOnCIter cit_grid = _grid->cbeginValueOn();
+  for (; cit_grid.test(); ++cit_grid) {
+
+    const openvdb::Coord pt_index(cit_grid.getCoord());
+    const openvdb::Vec3d pose_world = this->IndexToWorld(pt_index);
+
+    double x = pose_world[0];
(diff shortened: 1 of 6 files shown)


src/spatio_temporal_voxel_grid.cpp
```diff
@@ -451,4 +451,51 @@ bool SpatioTemporalVoxelGrid::SaveGrid(
...
+      if (this->ClearGridPoint(pt_index)) {
+        cleared_cells.insert(occupany_cell(x, y));
+      }
+      else {
+        std::cout << "Failed to clear point." << std::endl;
+      }
+    }
+    else {
+      // if here, we can add to costmap and PC2
+      PopulateCostmapAndPointcloud(pt_index);
```
SteveMacenski: Would this not already be happening via other mechanics?

ghost: That's true! it's done with the clearings on TemporalClearAndGenerateCostmap. Changed it so we only do it there, and also added restriction so it's only possible to clear when not on mapping mode, as it is for the frustum/temporal clearings.

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -768,6 +776,39 @@ void SpatioTemporalVoxelLayer::SaveGridCallback(
...
+}
+
+/*****************************************************************************/
+void SpatioTemporalVoxelLayer::ClearGridExceptRegion(
+  double robot_x, double robot_y,
+  std::unordered_set<volume_grid::occupany_cell> & cleared_cells)
+/*****************************************************************************/
+{
+  if (!_clear_grid_except_region)
+    return;
```
SteveMacenski, line 806: Add brackets in the if condition, its a 1 liners so I know you "can" but if we expanded that later I've had problems where developers forget to later add it

SteveMacenski (changes requested):
Last small change and good to go, I assume you tested this?

src/spatio_temporal_voxel_layer.cpp
ghost: Right, that's understandable. Added the brackets there. 

Yes have been testing and re-testing this after the changes!

SteveMacenski (approved):
This looks fine to me. If this were to be run often, I would rearchitect this so that we implement the local clearing in the actual ClearFrustum API so that we only iterate through the grid exactly 1 time, but since this is irregular and infrequent I think this is OK. If you do notice performance issues, that would be the low-hanging fruit solution. It would involve changing the API a little bit, but that way you can leverage the same iteration / world coordinate conversion that's already happening. 

its also probably possible in the openvdb API to define a range of values you're interested in and iterate only through those, but again, infrequently this is probably OK, even if suboptimal

Issue #224, opened 2022-02-09
There is an offset between the grid and the raw data


KavenYau:
Hi, I am trying to use STVL, but I found a issue that there is an offset between the grid of costmap and the raw data:
![image](https://user-images.githubusercontent.com/24326888/153141941-8d81ffcb-a6af-44f1-93d6-5a7f9b530c2f.png)

Did you see this before? And Have you any ideas of that?

More informations:

ROS version: Foxy
built from source: 396b25b806b22d9c256aa2000691f5dd171b4571
config of local_costmap:

```yaml
local_costmap:
  local_costmap:
    ros__parameters:
      update_frequency: 5.0
      publish_frequency: 2.0
      global_frame: odom
      robot_base_frame: base_link
      use_sim_time: True
      rolling_window: true
      width: 8
      height: 8
      resolution: 0.02
      footprint: "[[-0.32, -0.315], [-0.36, -0.275], [-0.36, 0.275], [-0.32, 0.315], [0.32, 0.315], [0.36, 0.275], [0.36, -0.275], [0.32, -0.315]]"
      footprint_padding: 0.01
      plugins: ["stvl_layer", "inflation_layer"]
      stvl_layer:
        plugin: "spatio_temporal_voxel_layer/SpatioTemporalVoxelLayer" # For Foxy and later
        enabled:                  true
        voxel_decay:              1.0  # seconds if linear, e^n if exponential
        decay_model:              0     # 0=linear, 1=exponential, -1=persistent
        voxel_size:               0.04  # meters
        track_unknown_space:      true  # default space is known
        max_obstacle_height:      2.0   # meters
        unknown_threshold:        15    # voxel height
        mark_threshold:           0     # voxel height
        update_footprint_enabled: true
        combination_method:       1     # 1=max, 0=override
        origin_z:                 0.0   # meters
        publish_voxel_map:        false # default off
        transform_tolerance:      0.2   # seconds
        mapping_mode:             false # default off, saves map not for navigation
        map_save_duration:        60.0  # default 60s, how often to autosave

        observation_sources:  scan
        scan:
          topic: /scan
          max_obstacle_height: 2.0
          clearing: True
          marking: True
          data_type: "LaserScan"
          raytrace_max_range: 15.0
          raytrace_min_range: 0.0
          obstacle_max_range: 15.0
          obstacle_min_range: 0.0
```

SteveMacenski:
[...]

PR #226, opened 2022-02-16
Fixing the offset that exists between the grid and the raw data

KavenYau:
Fixed https://github.com/SteveMacenski/spatio_temporal_voxel_layer/issues/224.

Without this patch:
image

With this patch:
image

### DIFF
--- a/src/spatio_temporal_voxel_grid.cpp
+++ b/src/spatio_temporal_voxel_grid.cpp
@@ -225,7 +225,7 @@ void SpatioTemporalVoxelGrid::PopulateCostmapAndPointcloud(
 /*****************************************************************************/
 {
   // add pt to the pointcloud and costmap
-  openvdb::Vec3d pose_world = _grid->indexToWorld(pt);
+  openvdb::Vec3d pose_world = this->IndexToWorld(pt);
 
   if (_pub_voxels) {
     geometry_msgs::msg::Point32 point;
@@ -285,8 +285,13 @@ void SpatioTemporalVoxelGrid::operator()(
       if (distance_2 > mark_range_2 || distance_2 < 0.0001) {
         continue;
       }
+
+      double x = *iter_x < 0 ? *iter_x - _voxel_size : *iter_x;
+      double y = *iter_y < 0 ? *iter_y - _voxel_size : *iter_y;
+      double z = *iter_y < 0 ? *iter_z - _voxel_size : *iter_z;
+
       openvdb::Vec3d mark_grid(this->WorldToIndex(
-          openvdb::Vec3d(*iter_x, *iter_y, *iter_z)));
+          openvdb::Vec3d(x, y, z)));
 
       if (!this->MarkGridPoint(openvdb::Coord(mark_grid[0], mark_grid[1],
         mark_grid[2]), cur_time))
@@ -412,7 +417,15 @@ openvdb::Vec3d SpatioTemporalVoxelGrid::IndexToWorld(
 /*****************************************************************************/
 {
   // Applies tranform stored in getTransform.
-  return _grid->indexToWorld(coord);
+  openvdb::Vec3d pose_world =  _grid->indexToWorld(coord);
+
+  // Using the center for world coordinate
+  const double & center_offset = _voxel_size / 2.0;
+  pose_world[0] += center_offset;
+  pose_world[1] += center_offset;
+  pose_world[2] += center_offset;
+
+  return pose_world;
 }
 
 /*****************************************************************************/


src/spatio_temporal_voxel_grid.cpp
```diff
@@ -412,7 +417,15 @@ openvdb::Vec3d SpatioTemporalVoxelGrid::IndexToWorld(
 /*****************************************************************************/
 {
   // Applies tranform stored in getTransform.
-  return _grid->indexToWorld(coord);
+  openvdb::Vec3d pose_world =  _grid->indexToWorld(coord);
```
SteveMacenski, line 420: This is not the approved solution in https://github.com/SteveMacenski/spatio_temporal_voxel_layer/commit/7bf1e12aea72269cda30a483ddeefbe7a333e592.

SteveMacenski (changes requested):

src/spatio_temporal_voxel_grid.cpp
SteveMacenski: Why not use that one? There's no reason to mess with the internals of this if we can simply apply an offset when converting to grid coordinates for the costmap

KavenYau: Because https://github.com/SteveMacenski/spatio_temporal_voxel_layer/commit/7bf1e12aea72269cda30a483ddeefbe7a333e592 did not work fine.
image

I noticed that if the pose_world represents the vertex of the grid, the grid occupation depends on different quadrants. I guess it's caused by the precision of double value, and there is deviation when converting to grid coordinates for the costmap. So I apply an offset to all points so that the pose_world represents the center of the grid. It works fine when costmap resolution >= voxel_size.

when costmap resolution(0.02) < voxel_size(0.04): 
image

SteveMacenski: OK, let me take a look at your solution then next week if we should be using this instead so I can update all of the appropriate locations in other branches as well.

SteveMacenski (approved):
I approve of this, but I would like @doisyg to review and also approve as I know he is a user of this and I don't want to break his behavior. Plus he worked on the initial alignment bits so he would be a good person to review this as well.

@doisyg can you review this PR?

If he approves, I'll backport / foward port to all distributions. This would override the changes from the other method used in `ros2`.

SteveMacenski:
My sincerest apologies for the delay, this really wasn't that complicated and I kept pushing it off.

doisyg:
Ok, quickly jumping in, I am a bit confused. Is this supposed to be only porting to Foxy the fix I did for Galactic there : https://github.com/SteveMacenski/spatio_temporal_voxel_layer/pull/213
or there is something more than a visualization issue to fix?

EDIT: reading the full conversation

SteveMacenski:
I think what the author has pointed out is that the previous solution didn’t actually fully fix the issue and this does. 

While this is targeted at Foxy, if you agree with these changes, I would update all ROS2 branches with this change and revert the previous fix. 

While visualization is part of it, if the grid is offset to the costmap, then that would have a real world impact of a half cell shift of measurements vs costmap cells occupied potentially

doisyg:
> I think what the author has pointed out is that the previous solution didn’t actually fully fix the issue and this does.

Yes I just read the comments.

There is maybe an issue with my fix when costmap_resoltion != voxel_size, but not sure this PR completely fixes it. I would like to take the time to run a couple of tests. I willl post here in the following days comparative results with different conditions and code version.
In the meantime, @KavenYau, can you clarify a bit more what you believe the original issue is and in which conditions https://github.com/SteveMacenski/spatio_temporal_voxel_layer/pull/213 is not enough ? I propose we keep the following convention when displaying results (as in https://github.com/SteveMacenski/spatio_temporal_voxel_layer/pull/213 ):

- lethal pixels are in pink
- input stvl sensor pointcloud is the small blue spheres 
- voxel visualization is in green and displayed as boxes with alpha 0.4 and of size equal to pixel size 
- nothing else is displayed (i.e. no inflation)
- view is TopDownOrtho to avoid false inference due to perspective
- zoomed enough to clearly see misalignement

For instance the following example shows misalignment of voxel visualization (size 0.05) against the costmap grid and the input pointcloud:
image

KavenYau:
> can you clarify a bit more what you believe the original issue is and in which conditions https://github.com/SteveMacenski/spatio_temporal_voxel_layer/pull/213 is not enough ?

I think there are 3 points:
1. I guess tests in #213 did not cover all quadrants. 
2. #213 is assuming the lethal pixels are correct. But I made a small [experiment](https://github.com/KavenYau/openvdb_demo.git) and the result shows that it's not, all the numbers in one of the directions(positive or negative) on each axis have an offset. Described in https://github.com/SteveMacenski/spatio_temporal_voxel_layer/issues/224#issuecomment-1040951516.
3. As #213 mentioned, when costmap_resoltion is equals to voxel_size, the pointcloud where the x and y coordinates of each point are the coordinates of the corner of the voxel. Below figure shows that the lethal grids of costmap tend to extend towards the origin ( like yellow arrows).
image

> I propose we keep the following convention when displaying results (as in https://github.com/SteveMacenski/spatio_temporal_voxel_layer/pull/213 ):

Ok, but currently I haven't the environment. I will create it if necessary.

doisyg:
Sorry for the late reply, I was busy and wanted to do some proper testing on a real hardware.
I can confirm that https://github.com/SteveMacenski/spatio_temporal_voxel_layer/pull/213 is not fixing the issue when  `costmap_resoltion != voxel_size` (current galactic branch)
I did some tests on the galactic branch (my HW environment) with a revert of https://github.com/SteveMacenski/spatio_temporal_voxel_layer/pull/213 and by applying this PR. It seems to fix the issue for all cases. Tests done with a sensor generating pointclouds (and not laserscan) with a limited fov, hence only on the 2 forward quadrant.
I did not do more digging by this looks good to me. Nice catch @KavenYau !

SteveMacenski:
OK - just to be clear before I do this, you're OK if I:
- Revert #213 for all branches
- Apply this for all branches

doisyg:
Yes !

Issue #232, opened 2022-04-24
Impossible to build for Ubuntu Jammy 22.04 / Humble (probably because of OpenVDB / TBB)


doisyg:
Hello,
FYI and to track the issue:
In preparation for the Humble release I tried to build STVL under Humble beta with Ubuntu 22.04 and the blocking point seems to be, surprise, the version of the OpenVDB package released in Ubuntu.
My quick analysis of the cause:

- OpenVDB was upgraded from **v6.2** in `Ubuntu 20.04/focal` to **v8.1** in `Ubuntu 22.04/jammy`
- TBB was upgraded from **2020.1** in `Ubuntu 20.04/focal` to **2021.5** in `Ubuntu 22.04/jammy`
- OpenVDB 8.1 uses headers from TBB 2020 which were deprecated a while ago and now are completely gone in **TBB 2021.5**. For instance `#include <tbb/task_scheduler_init.h>` https://github.com/AcademySoftwareFoundation/openvdb/blob/ea786c46b7a1b5158789293d9b148b379fc9914c/openvdb/openvdb/tools/VolumeToMesh.h#L21
See https://www.intel.com/content/www/us/en/developer/articles/release-notes/intel-oneapi-threading-building-blocks-release-notes.html and https://github.com/DrTimothyAldenDavis/SuiteSparse/issues/72

OpenVDB seems to already have a version (**v8.2**) compatible with **TBB 2021.5** since november : https://github.com/AcademySoftwareFoundation/openvdb/releases/tag/v8.2.0. My guess is that releasing **v8.2** instead of **v8.1** to `Ubuntu 22.04/jammy` would solve the issue. I will try to file a bug report on launchpad hoping that it will be faster than https://bugs.launchpad.net/ubuntu/+source/openvdb/+bug/1882998 (remember https://github.com/SteveMacenski/spatio_temporal_voxel_layer/issues/167).

(Side note, looking at OpenVDB release page: https://github.com/AcademySoftwareFoundation/openvdb/releases, I noticed the release of v9.0 with 

> Official release of NanoVDB, which for the first time offers GPU support for static sparse volumes in OpenVDB.

It is GPU agnotic, more info: https://academysoftwarefoundation.github.io/openvdb/NanoVDB_FAQ.html
@SteveMacenski, did you follow ? Maybe that's interesting to look at for STVL and voxels in nav2)

SteveMacenski:
[...]

PR #233, opened 2022-04-28
Humble build

doisyg:
Changes for Humble compilation.
Current limitation: Ubuntu provided `OpenVdb 8.1` issues, see: https://github.com/SteveMacenski/spatio_temporal_voxel_layer/issues/232

Works by uninstalling `libopenvdb-dev` and building/installing OpenVDB v8.2 (also tried successfully v9.0) from source with these instructions https://github.com/AcademySoftwareFoundation/openvdb#building-openvdb

Draft because not tested for now (just compilation)

### DIFF
--- a/src/spatio_temporal_voxel_layer.cpp
+++ b/src/spatio_temporal_voxel_layer.cpp
@@ -831,7 +831,7 @@ SpatioTemporalVoxelLayer::dynamicParametersCallback(std::vector<rclcpp::Paramete
     while (ss >> source) {
       if (type == ParameterType::PARAMETER_DOUBLE) {
         if (name == name_ + "." + source + "." + "min_obstacle_height") {
-          for (auto & buffer:_observation_buffers) {
+          for (auto & buffer : _observation_buffers) {
             if (buffer->GetSourceName() == source) {
               buffer->Lock();
               buffer->SetMinObstacleHeight(parameter.as_double());
@@ -839,7 +839,7 @@ SpatioTemporalVoxelLayer::dynamicParametersCallback(std::vector<rclcpp::Paramete
             }
           }
         } else if (name == name_ + "." + source + "." + "max_obstacle_height") {
-          for (auto & buffer:_observation_buffers) {
+          for (auto & buffer : _observation_buffers) {
             if (buffer->GetSourceName() == source) {
               buffer->Lock();
               buffer->SetMaxObstacleHeight(parameter.as_double());
@@ -847,7 +847,7 @@ SpatioTemporalVoxelLayer::dynamicParametersCallback(std::vector<rclcpp::Paramete
             }
           }
         } else if (name == name_ + "." + source + "." + "min_z") {
-          for (auto & buffer:_observation_buffers) {
+          for (auto & buffer : _observation_buffers) {
             if (buffer->GetSourceName() == source) {
               buffer->Lock();
               buffer->SetMinZ(parameter.as_double());
@@ -855,7 +855,7 @@ SpatioTemporalVoxelLayer::dynamicParametersCallback(std::vector<rclcpp::Paramete
             }
           }
         } else if (name == name_ + "." + source + "." + "max_z") {
-          for (auto & buffer:_observation_buffers) {
+          for (auto & buffer : _observation_buffers) {
             if (buffer->GetSourceName() == source) {
               buffer->Lock();
               buffer->SetMaxZ(parameter.as_double());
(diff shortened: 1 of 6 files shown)


src/spatio_temporal_voxel_layer.cpp
```diff
@@ -329,8 +339,8 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
       _1, _2, _3, _observation_buffers.back(),
       _observation_subscribers.back());
     std::string toggle_topic = source + "/toggle_enabled";
-    auto server = rclcpp_node_->create_service<std_srvs::srv::SetBool>(
-      toggle_topic, toggle_srv_callback);
+    auto server = node->create_service<std_srvs::srv::SetBool>(
+      toggle_topic, toggle_srv_callback, rmw_qos_profile_services_default, callback_group_);
```
gezp, line 343: Same as above.

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -140,13 +140,19 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
...
+
+  _voxel_pub = node->create_publisher<sensor_msgs::msg::PointCloud2>(
+    "voxel_grid", rclcpp::QoS(1), pub_opt);
 
   auto save_grid_callback = std::bind(
     &SpatioTemporalVoxelLayer::SaveGridCallback, this, _1, _2, _3);
-  _grid_saver = rclcpp_node_->create_service<spatio_temporal_voxel_layer::srv::SaveGrid>(
-    "save_grid", save_grid_callback);
+  _grid_saver = node->create_service<spatio_temporal_voxel_layer::srv::SaveGrid>(
+    "save_grid", save_grid_callback, rmw_qos_profile_services_default, callback_group_);
```
gezp, line 155: Same as above.

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -140,13 +140,19 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
...
-  _voxel_pub = rclcpp_node_->create_publisher<sensor_msgs::msg::PointCloud2>(
-    "voxel_grid", rclcpp::QoS(1));
+  auto sub_opt = rclcpp::SubscriptionOptions();
+  sub_opt.callback_group = callback_group_;
+
+  auto pub_opt = rclcpp::PublisherOptions();
+  sub_opt.callback_group = callback_group_;
+
+  _voxel_pub = node->create_publisher<sensor_msgs::msg::PointCloud2>(
+    "voxel_grid", rclcpp::QoS(1), pub_opt);
```
gezp, line 150: it's equivalent behavior as before, but i'm not sure why use `rclcpp_node_` to create `_voxel_pub` before, maybe it's unnecessary.

currently, `callback_group_` in `nav2_costmap_2d::Layer` was only used for tf message filter in navigation2 repo .

by the way, this is a discussion, you can ignore it if you think we need keep equivalent as before.

SteveMacenski:
Beyond comments above, just linting spaces between varables / definitions and `&`  and removing a commented out line in the header. 

Note that you should try on Cyclone DDS, for Fast-DDS right now callback groups are broke so data won't flow.

I don't think the pub options are required, but I suppose it doesn't hurt.

src/spatio_temporal_voxel_layer.cpp
doisyg: > it's equivalent behavior as before, but i'm not sure why use `rclcpp_node_` to create `_voxel_pub` before, maybe it's unnecessary.
> 
> currently, `callback_group_` in `nav2_costmap_2d::Layer` was only used for tf message filter in navigation2 repo .

Thanks for the comments, as soon as I can try it on real hardware, I will see if it works without the callback group. Leaving it for now for equivalency.
Something that I am not 100% sure also is if the service are going to work. Do you have any example of working services in a layer since your PR ?

doisyg:
> Note that you should try on Cyclone DDS, for Fast-DDS right now callback groups are broke so data won't flow.

Of gosh I know, don't get me started on Fast-DDS.

In other news, the launchpad ticket about OpenVDB status changed form `New` to `Confirmed`: https://bugs.launchpad.net/ubuntu/+source/openvdb/+bug/1970108

SteveMacenski:
Ok, I won't :-) 

I think this PR is waiting on you for the next step to test / change? Just making sure I'm not the blocking force. No rush.

doisyg:
This was waiting for me to do some basic sanity checks on a real robot. Done now.

SteveMacenski (approved):

SteveMacenski:
I'm happy with this if you've tested. @gezp @doisyg any reason not to merge?

doisyg:
All good on my side given the basic testing I did.

gezp:
LGTM

PR #243, opened 2022-12-13
Implement CostmapLayer::clearArea

siferati:
## Description

- Added custom implementation of `CostmapLayer::clearArea` since the [default implementation](https://github.com/ros-planning/navigation/blob/4122c1b84296b7c49799ef1aee742565090753d9/costmap_2d/src/costmap_layer.cpp#L21-L36) doesn't work for STVL.
- I guess this closes #122?
- **Bonus:** some minor whitespace clearing performed by vscode.

## Demo

You can see in the video the STVL layer getting cleared as I call `ClearCostmapRecovery` behavior.

https://user-images.githubusercontent.com/15384781/207238088-f119a664-969f-45e4-bb7c-d461cc8c69a8.mp4


corot (approved):
Works fine; great contribution

SteveMacenski (approved):
Can you open a complimentary PR for `ros2` and then I'll merge the pair? Just keeps this all in sync.

Great to see contributions from Rapyuta! :smile:

siferati:
@SteveMacenski I opened the PR but I do not have a ROS2 system to test on so would be nice if you could make sure it at least compiles before merging!

SteveMacenski (approved):

SteveMacenski:
Thanks!

corot:
@SteveMacenski, I think this justifies a release, as before we where essentially ignoring clear_costmap recovery behavior :thinking:    would you mind?

SteveMacenski:
Will do

corot:
> Will do

wait, wait,,,,  we found an issue with the mutex locking that can freeze move_base_flex

I think I found the issue: https://github.com/rapyuta-robotics/spatio_temporal_voxel_layer/pull/1
@SteveMacenski , can u take a look and decide which option is the best?
- add the costmap lock (current PR implementation)
- remove the voxel lock on clearArea

PR #252, opened 2023-02-18
Fix "enable" parameter for observation sources

HovorunB:
I noticed that if we set `enabled: false` for [observation sources](https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/94e8439f7d1ff7038ce9c1138df5d2e7c743dbc2/src/spatio_temporal_voxel_layer.cpp#L239) it doesn't actually disable them on startup.
I think the problem is that `enabled` parameter is responsible for creating / destroying subscription only on service call, and is not considered for the subscription creation on startup. Or am I missing smth?

In this PR I propose to add checking for `enabled` parameter on startup to make sure that the observation source is not subsrcibed to the topic if it is disabled

Another approach could be to use [this implementation](https://github.com/ros2/message_filters/blob/de46fa8086f4b5586ea3ea80fcbbce8984ad15d2/include/message_filters/subscriber.h#L203) 
In short:
Create subscription objects without actually subscribing to the topic, then subscribe to the topics from `activate` function and only for observation sources which are enabled

Wanted to get your thoughts about implementation. Maybe you have other ideas?

### DIFF
--- a/src/spatio_temporal_voxel_layer.cpp
+++ b/src/spatio_temporal_voxel_layer.cpp
@@ -552,6 +552,7 @@ void SpatioTemporalVoxelLayer::activate(void)
     }
     (*buf_it)->ResetLastUpdatedTime();
   }
+
   // Add callback for dynamic parametrs
   auto node = node_.lock();
   dyn_params_handler = node->add_on_set_parameters_callback(
--- a/src/spatio_temporal_voxel_layer.cpp
+++ b/src/spatio_temporal_voxel_layer.cpp
@@ -552,7 +552,6 @@ void SpatioTemporalVoxelLayer::activate(void)
     }
     (*buf_it)->ResetLastUpdatedTime();
   }
-
   // Add callback for dynamic parametrs
   auto node = node_.lock();
   dyn_params_handler = node->add_on_set_parameters_callback(
--- a/src/spatio_temporal_voxel_layer.cpp
+++ b/src/spatio_temporal_voxel_layer.cpp
@@ -545,13 +545,15 @@ void SpatioTemporalVoxelLayer::activate(void)
   RCLCPP_INFO(logger_, "%s was activated.", getName().c_str());
 
   observation_subscribers_iter sub_it = _observation_subscribers.begin();
+  for (; sub_it != _observation_subscribers.end(); ++sub_it) {
+    (*sub_it)->subscribe();
+  }
+
   observation_buffers_iter buf_it = _observation_buffers.begin();
-  for (; sub_it != _observation_subscribers.end(); ++sub_it, ++buf_it) {
-    if (!(*buf_it)->IsEnabled()) {
-      (*sub_it)->unsubscribe();
-    }
+  for (; buf_it != _observation_buffers.end(); ++buf_it) {
     (*buf_it)->ResetLastUpdatedTime();
   }
+
   // Add callback for dynamic parametrs
   auto node = node_.lock();
   dyn_params_handler = node->add_on_set_parameters_callback(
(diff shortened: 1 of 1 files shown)


src/spatio_temporal_voxel_layer.cpp
```diff
@@ -545,12 +545,11 @@ void SpatioTemporalVoxelLayer::activate(void)
   RCLCPP_INFO(logger_, "%s was activated.", getName().c_str());
 
   observation_subscribers_iter sub_it = _observation_subscribers.begin();
-  for (; sub_it != _observation_subscribers.end(); ++sub_it) {
-    (*sub_it)->subscribe();
```
SteveMacenski: Why did you remove the subscribe function? We need this to start processing things for any topics at all regardless of enabled status.

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -545,12 +545,11 @@ void SpatioTemporalVoxelLayer::activate(void)
...
-  }
-
   observation_buffers_iter buf_it = _observation_buffers.begin();
-  for (; buf_it != _observation_buffers.end(); ++buf_it) {
+  for (; sub_it != _observation_subscribers.end(); ++sub_it, ++buf_it) {
+    if (!(*buf_it)->IsEnabled()) {
```
SteveMacenski: buf it is not set anywhere of anything except `begin()`

SteveMacenski (changes requested):
This PR has functional issues.

src/spatio_temporal_voxel_layer.cpp
HovorunB: When the subscriber is created, it automatically subscribes to the topic because we pass all the needed parameters, see:
https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/94e8439f7d1ff7038ce9c1138df5d2e7c743dbc2/src/spatio_temporal_voxel_layer.cpp#L315
and
https://github.com/ros2/message_filters/blob/de46fa8086f4b5586ea3ea80fcbbce8984ad15d2/include/message_filters/subscriber.h#L182

src/spatio_temporal_voxel_layer.cpp
HovorunB: it is set from the `toggle` service [callback ](https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/94e8439f7d1ff7038ce9c1138df5d2e7c743dbc2/src/spatio_temporal_voxel_layer.cpp#L444)
I think this was initially the desired functionality, the `enable` parameter can disable the subscription on startup, and then it can subscribe by calling the service. [docstring](https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/94e8439f7d1ff7038ce9c1138df5d2e7c743dbc2/src/spatio_temporal_voxel_layer.cpp#L238)

SteveMacenski: The orders of sub_it and buf_its are not guaranteed to be aligned if the configuration is malformed.

src/spatio_temporal_voxel_layer.cpp
SteveMacenski: And what happens if you deactivate the system to reactivate it? :smile:

HovorunB:
Sorry for the delay, at first i tried to solve this in a way how (under my assumption) it was done before (when creating a subscription object didn't automatically subscribe to the topic). But now i see that this might not be a very good solution. Then I suggest another one, to check if observation source in enabled on callback, and return otherwise. This way we don't rely on length of sub_it and buf_its, and keep the logic on reactivation

SteveMacenski (approved):

PR #259, opened 2023-03-29
add 'enabled' parameter to set dynamically

dujeong:
There was already a service that had a function similar to toggle stvl. (service name: `<observation_topic>/toggle_enabled`)
The service can be set up for each individual topic, but I wanted a toggle function for the entire stvl.

So I implemented the function and made it dynamically configurable.

I referred to the functions(activate, deactivate).

### DIFF
--- a/src/spatio_temporal_voxel_layer.cpp
+++ b/src/spatio_temporal_voxel_layer.cpp
@@ -902,6 +902,33 @@ SpatioTemporalVoxelLayer::dynamicParametersCallback(std::vector<rclcpp::Paramete
         }
       }
     }
+
+    if (type == ParameterType::PARAMETER_BOOL) {
+      if (name == name_ + "." + "enabled") {
+        bool enable = parameter.as_bool();
+        if (enabled_ != enable) {
+          if (enable){
+            observation_subscribers_iter sub_it = _observation_subscribers.begin();
+            for (; sub_it != _observation_subscribers.end(); ++sub_it) {
+              (*sub_it)->subscribe();
+            }
+
+            observation_buffers_iter buf_it = _observation_buffers.begin();
+            for (; buf_it != _observation_buffers.end(); ++buf_it) {
+              (*buf_it)->ResetLastUpdatedTime();
+            }
+          } else {
+            observation_subscribers_iter sub_it = _observation_subscribers.begin();
+            for (; sub_it != _observation_subscribers.end(); ++sub_it) {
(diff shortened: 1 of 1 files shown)


src/spatio_temporal_voxel_layer.cpp
```diff
@@ -902,6 +902,33 @@ SpatioTemporalVoxelLayer::dynamicParametersCallback(std::vector<rclcpp::Paramete
         }
       }
     }
+
+    if (type == ParameterType::PARAMETER_BOOL) {
+      if (name == name_ + "." + "enabled") {
+        bool enable = parameter.as_bool();
+        if (enabled_ != enable) {
+          if (enable){
```
SteveMacenski:
```suggestion
          if (enable) {
```

src/spatio_temporal_voxel_layer.cpp
```diff
@@ -902,6 +902,33 @@ SpatioTemporalVoxelLayer::dynamicParametersCallback(std::vector<rclcpp::Paramete
...
+            }
+
+            observation_buffers_iter buf_it = _observation_buffers.begin();
+            for (; buf_it != _observation_buffers.end(); ++buf_it) {
+              (*buf_it)->ResetLastUpdatedTime();
+            }
+          } else {
+            observation_subscribers_iter sub_it = _observation_subscribers.begin();
+            for (; sub_it != _observation_subscribers.end(); ++sub_it) {
+              if (*sub_it != NULL) {
```
SteveMacenski: Why do you check here if any are nullptrs but not above?

SteveMacenski (changes requested):
Generally looks good

src/spatio_temporal_voxel_layer.cpp
dujeong: It's because I copied and pasted...
Do you think the nullptr inspection part is necessary?
I don't think that part is necessary.
If you agree, should I also remove it from the "deactivate" function?

SteveMacenski: Add it here

dujeong:
@SteveMacenski 
I added checking nullptr part in the code. Is that what you meant?

SteveMacenski (approved):
I assume you tested this as working?

dujeong:
Sure! I tested it on the robot. It worked well. :)

SteveMacenski:
Thanks for the contribution!

PR #264, opened 2023-06-02
Allow mark_threshold parameter to be set dynamically

HovorunB:
Hello! :slightly_smiling_face: In my usecase sometimes I need to disable marking the costmap, but keep publishing STVL to the topic, and reconfiguring mark_threshold is the way to go for me.
So if you think it could be useful for somebody else then I propose to add the ability to set dynamically `mark_threshold` parameter


SteveMacenski (approved):

SteveMacenski:
Sounds good!

PR #266, opened 2023-07-13
Kinetic/fix/vertical fov bug

yShzZpp:
Subject: Pull Request: Modification for spatioTemporalVoxelLayer

Dear Steve,

I hope this message finds you well. I am submitting my first Pull Request for your repository, [Repository Name], and I would greatly appreciate your review and consideration. I have made modifications to the code related to the usage of the spatioTemporalVoxelLayer, as I encountered compatibility issues with the kinetic version of the library when using a Livox radar.

The Livox radar has a field of view (FOV) ranging from -7° to 53°. However, the existing code simply divides the FOV by two as a parameter, resulting in an incomplete clearance of the costmap in the upper half, while the lower half is cleared properly.

To address this issue, I have made the following modifications to the code:

Replaced the "vertical_fov_angle" parameter with two parameters: "vertical_fov_start_angle" and "vertical_fov_end_angle." This allows for accurate definition of the vertical FOV range and resolves the clearance issue.
I have thoroughly tested these modifications on the Livox radar. However, I must note that I do not have access to a depth camera for comprehensive testing. Therefore, I have only submitted the modifications for the 3D radar functionality.

I genuinely value your expertise and would greatly appreciate your feedback, suggestions, or any optimizations you may have. This Pull Request holds significant meaning for my professional career, and I am committed to making any necessary changes to ensure its successful integration into the project.

Thank you for taking the time to review my code, despite your busy schedule. I look forward to your response.

Best regards,
zzpp

### DIFF
--- a/include/spatio_temporal_voxel_layer/frustum_models/depth_camera_frustum.hpp
+++ b/include/spatio_temporal_voxel_layer/frustum_models/depth_camera_frustum.hpp
@@ -53,7 +53,7 @@ namespace geometry
 class DepthCameraFrustum : public Frustum
 {
 public:
-  DepthCameraFrustum(const double& vSFOV, const double& vEFOV, const double& hFOV,
+  DepthCameraFrustum(const double& vFOV, const double& hFOV,
           const double& min_dist, const double& max_dist);
   virtual ~DepthCameraFrustum(void);
 
@@ -73,7 +73,7 @@ private:
   double Dot(const VectorWithPt3D&, const openvdb::Vec3d&) const;
   double Dot(const VectorWithPt3D&, const Eigen::Vector3d&) const;
 
-  double _vSFOV, _hFOV, _min_d, _max_d;
+  double _vFOV, _hFOV, _min_d, _max_d;
   std::vector<VectorWithPt3D> _plane_normals;
   Eigen::Vector3d _position;
   Eigen::Quaterniond _orientation;
--- a/include/spatio_temporal_voxel_layer/frustum_models/depth_camera_frustum.hpp
+++ b/include/spatio_temporal_voxel_layer/frustum_models/depth_camera_frustum.hpp
@@ -53,7 +53,7 @@ namespace geometry
 class DepthCameraFrustum : public Frustum
 {
 public:
-  DepthCameraFrustum(const double& vFOV, const double& hFOV,
+  DepthCameraFrustum(const double& vSFOV, const double& vEFOV, const double& hFOV,
           const double& min_dist, const double& max_dist);
   virtual ~DepthCameraFrustum(void);
 
@@ -73,7 +73,7 @@ private:
   double Dot(const VectorWithPt3D&, const openvdb::Vec3d&) const;
   double Dot(const VectorWithPt3D&, const Eigen::Vector3d&) const;
 
-  double _vFOV, _hFOV, _min_d, _max_d;
+  double _vSFOV, _hFOV, _min_d, _max_d;
   std::vector<VectorWithPt3D> _plane_normals;
   Eigen::Vector3d _position;
   Eigen::Quaterniond _orientation;
--- a/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
+++ b/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
@@ -52,7 +52,7 @@ namespace geometry
 class ThreeDimensionalLidarFrustum : public Frustum
 {
 public:
-  ThreeDimensionalLidarFrustum(const double& vFOV, const double& vFOVPadding, 
+  ThreeDimensionalLidarFrustum(const double& vEFOV,const double& vSFOV, const double& vFOVPadding, 
           const double& hFOV, const double& min_dist, const double& max_dist);
   virtual ~ThreeDimensionalLidarFrustum(void);
 
@@ -71,11 +71,11 @@ private:
   double Dot(const VectorWithPt3D&, const openvdb::Vec3d&) const;
   double Dot(const VectorWithPt3D&, const Eigen::Vector3d&) const;
 
-  double _vFOV, _vFOVPadding, _hFOV, _min_d, _max_d;
+  double _vSFOV, _vEFOV, _vFOVPadding, _hFOV, _min_d, _max_d;
   double _hFOVhalf;
   double _min_d_squared, _max_d_squared;
-  double _tan_vFOVhalf;
-  double _tan_vFOVhalf_squared;
+  double _tan_vSFOV, _tan_vEFOV;
+  double _tan_vSFOV_squared, _tan_vEFOV_squared;
   Eigen::Vector3d _position;
   Eigen::Quaterniond _orientation;
   Eigen::Quaterniond _orientation_conjugate;
--- a/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
+++ b/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
@@ -52,8 +52,9 @@ namespace geometry
 class ThreeDimensionalLidarFrustum : public Frustum
 {
 public:
-  ThreeDimensionalLidarFrustum(const double& vFOV, const bool& use_start_end_angle, const double& vEFOV,const double& vSFOV,
-          const double& vFOVPadding, const double& hFOV, const double& min_dist, const double& max_dist);
+  ThreeDimensionalLidarFrustum(const double& vFOV, const bool& use_start_end_angle,
+               const double& vEFOV, const double& vSFOV, const double& vFOVPadding,
+               const double& hFOV, const double& min_dist, const double& max_dist);
   virtual ~ThreeDimensionalLidarFrustum(void);
 
   // Does nothing in 3D lidar model
--- a/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
+++ b/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
@@ -52,10 +52,8 @@ namespace geometry
 class ThreeDimensionalLidarFrustum : public Frustum
 {
 public:
-  ThreeDimensionalLidarFrustum(const double& vFOV, const double& vFOVPadding, 
-          const double& hFOV, const double& min_dist, const double& max_dist);
-  ThreeDimensionalLidarFrustum(const double& vEFOV,const double& vSFOV, const double& vFOVPadding,
-          const double& hFOV, const double& min_dist, const double& max_dist);
+  ThreeDimensionalLidarFrustum(const double& vFOV, const bool& use_start_end_angle, const double& vEFOV,const double& vSFOV,
+          const double& vFOVPadding, const double& hFOV, const double& min_dist, const double& max_dist);
   virtual ~ThreeDimensionalLidarFrustum(void);
 
   // Does nothing in 3D lidar model
@@ -73,7 +71,7 @@ private:
   double Dot(const VectorWithPt3D&, const openvdb::Vec3d&) const;
   double Dot(const VectorWithPt3D&, const Eigen::Vector3d&) const;
 
-  bool _use_start_end_fov;
+  bool _use_start_end_angle;
   double _vFOV, _vSFOV, _vEFOV, _vFOVPadding, _hFOV, _min_d, _max_d;
   double _hFOVhalf;
   double _min_d_squared, _max_d_squared;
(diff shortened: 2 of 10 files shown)


include/spatio_temporal_voxel_layer/frustum_models/depth_camera_frustum.hpp
```diff
@@ -73,7 +73,7 @@ class DepthCameraFrustum : public Frustum
   double Dot(const VectorWithPt3D&, const openvdb::Vec3d&) const;
   double Dot(const VectorWithPt3D&, const Eigen::Vector3d&) const;
 
-  double _vFOV, _hFOV, _min_d, _max_d;
+  double _vSFOV, _hFOV, _min_d, _max_d;
```
SteveMacenski: Why is the "E" variant not stored?

src/frustum_models/depth_camera_frustum.cpp
```diff
@@ -41,9 +41,9 @@ namespace geometry
 {
 
 /*****************************************************************************/
-DepthCameraFrustum::DepthCameraFrustum(const double& vFOV, const double& hFOV,
+DepthCameraFrustum::DepthCameraFrustum(const double& vSFOV, const double& vEFOV, const double& hFOV,
                               const double& min_dist, const double& max_dist) :
-                   _vFOV(vFOV), _hFOV(hFOV), _min_d(min_dist), _max_d(max_dist)
+                   _vSFOV(vSFOV), _hFOV(hFOV), _min_d(min_dist), _max_d(max_dist)
```
SteveMacenski: Why is "E" not stored?

src/frustum_models/depth_camera_frustum.cpp
```diff
@@ -81,18 +81,18 @@ void DepthCameraFrustum::ComputePlaneNormals(void)
 
   // rotate going CCW
   Eigen::Affine3d rx =
-    Eigen::Affine3d(Eigen::AngleAxisd(_vFOV/2.,Eigen::Vector3d::UnitX()));
```
SteveMacenski: Why is E never used?

SteveMacenski (changes requested):
I think it would be wise to have this as an 'in addition to' feature instead of 'in place of'. The vast majority of users don't have asymmetric frustrums so having the option to continue to use the current API is preferable.

yShzZpp:
Dear Steve,
In this latest revision, I have made changes to incorporate FOV segmentation for 3D LiDAR scans. The segmentation will only be enabled when "model_type = 1" and "use_start_end_angle = true". For cases involving a depth camera, this feature will remain inactive.
Best regards,
zzpp

include/spatio_temporal_voxel_layer/measurement_buffer.hpp
```diff
@@ -71,7 +71,7 @@ typedef pcl::PointCloud<pcl::PointXYZ>::Ptr point_cloud_ptr;
 class MeasurementBuffer
 {
 public:
-  MeasurementBuffer(const std::string& topic_name,          \
+	MeasurementBuffer(const std::string& topic_name,          \
```
SteveMacenski: Revert the spacing change here

src/frustum_models/three_dimensional_lidar_frustum.cpp
```diff
@@ -65,6 +66,36 @@ ThreeDimensionalLidarFrustum::ThreeDimensionalLidarFrustum(const double& vFOV,
   }
 }
 
+/*****************************************************************************/
+ThreeDimensionalLidarFrustum::ThreeDimensionalLidarFrustum(const double& vSFOV,
```
SteveMacenski: Is the new contructor really needed?

SteveMacenski (changes requested):
Much better!

yShzZpp:
Dear Steve,
I wanted to express my gratitude for your patient guidance and support during the review process. I have made further adjustments to the code.
In this revised version, I have implemented a method that utilizes only a single constructor. I carefully considered your feedback and worked diligently to ensure that the code adheres to the best practices and standards of the project.
I kindly request you to review this latest iteration, and if there are any lingering issues or concerns, please do not hesitate to point them out. I am fully committed to patiently continue refining the code until it meets your expectations.

Once again, I extend my heartfelt appreciation for your valuable input and guidance. Your support is pivotal in my journey of growth and improvement as a developer.

Thank you for taking the time to review my work. I look forward to your feedback.

Best regards,
zzpp

include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
```diff
@@ -52,8 +52,8 @@ namespace geometry
 class ThreeDimensionalLidarFrustum : public Frustum
 {
 public:
-  ThreeDimensionalLidarFrustum(const double& vFOV, const double& vFOVPadding, 
-          const double& hFOV, const double& min_dist, const double& max_dist);
+  ThreeDimensionalLidarFrustum(const double& vFOV, const bool& use_start_end_angle, const double& vEFOV,const double& vSFOV,
```
SteveMacenski:
```suggestion
  ThreeDimensionalLidarFrustum(const double& vFOV, const bool& use_start_end_angle, const double& vEFOV, const double& vSFOV,
```

SteveMacenski: I also doubt this line is < 100 lines

src/frustum_models/three_dimensional_lidar_frustum.cpp
```diff
@@ -100,8 +110,22 @@ bool ThreeDimensionalLidarFrustum::IsInside(const openvdb::Vec3d &pt)
...
+  // const double v_padded = fabs(transformed_pt[2]) + _vFOVPadding;
+  // if (( v_padded * v_padded / radial_distance_squared) > _tan_vSFOV_squared)
+  // {
+  //   return false;
+  // }
+  const double v_padded = transformed_pt[2] + _vFOVPadding;
+  double tan_vFOV_squared;
+  if (_use_start_end_angle)
+  {
+    tan_vFOV_squared = (v_padded < 1e-9 ? _tan_vSFOV_squared : _tan_vEFOV_squared);
```
SteveMacenski: Why not compare to `0`?

src/frustum_models/three_dimensional_lidar_frustum.cpp
```diff
@@ -100,8 +110,22 @@ bool ThreeDimensionalLidarFrustum::IsInside(const openvdb::Vec3d &pt)
...
 
   // // Check if inside frustum valid vFOV
-  const double v_padded = fabs(transformed_pt[2]) + _vFOVPadding;
-  if (( v_padded * v_padded / radial_distance_squared) > _tan_vFOVhalf_squared)
+  // const double v_padded = fabs(transformed_pt[2]) + _vFOVPadding;
+  // if (( v_padded * v_padded / radial_distance_squared) > _tan_vSFOV_squared)
+  // {
+  //   return false;
+  // }
+  const double v_padded = transformed_pt[2] + _vFOVPadding;
```
SteveMacenski: I don't think this necessarily works for the negative case, because the padding is always positive even when the transformed point coordinates are negative, that's why we have `fabs` above.

src/frustum_models/three_dimensional_lidar_frustum.cpp
```diff
@@ -100,8 +110,22 @@ bool ThreeDimensionalLidarFrustum::IsInside(const openvdb::Vec3d &pt)
...
+  double tan_vFOV_squared;
```
SteveMacenski:
```suggestion
  double tan_vFOV_squared = _tan_vFOVhalf_squared;
```
Then remove the else statement.

src/measurement_buffer.cpp
```diff
@@ -71,7 +74,9 @@ MeasurementBuffer::MeasurementBuffer(const std::string& topic_name,          \
     _topic_name(topic_name), _min_obstacle_height(min_obstacle_height), 
     _max_obstacle_height(max_obstacle_height), _obstacle_range(obstacle_range),
     _tf_tolerance(tf_tolerance), _min_z(min_d), _max_z(max_d), 
-    _vertical_fov(vFOV), _vertical_fov_padding(vFOVPadding),
+		_vertical_fov(vFOV), _use_start_end_angle(use_start_end_angle && model_type == ModelType::THREE_DIMENSIONAL_LIDAR),
```
SteveMacenski: misformatting.

src/measurement_buffer.cpp
```diff
@@ -71,7 +74,9 @@ MeasurementBuffer::MeasurementBuffer(const std::string& topic_name,          \
...
+    _vertical_start_fov(vSFOV),_vertical_end_fov(vEFOV),
```
SteveMacenski, line 79: misformatting

src/spatio_temporal_voxel_grid.cpp
```diff
@@ -132,12 +132,15 @@ void SpatioTemporalVoxelGrid::ClearFrustums(const \
...
     else if (it->_model_type == THREE_DIMENSIONAL_LIDAR)
     {
-      frustum = new geometry::ThreeDimensionalLidarFrustum( \
-                                                    it->_vertical_fov_in_rad,
-                                                    it->_vertical_fov_padding_in_m,
-                                                    it->_horizontal_fov_in_rad,
-                                                    it->_min_z_in_m,
-                                                    it->_max_z_in_m);
+        frustum = new geometry::ThreeDimensionalLidarFrustum( \
+                                                      it->_vertical_fov_in_rad,
```
SteveMacenski: Misformating

SteveMacenski (changes requested):

src/measurement_buffer.cpp
yShzZpp: I'm sorry, I'm not sure what the problem is here, could you please answer it? Or suggest a change

src/spatio_temporal_voxel_grid.cpp
yShzZpp: I'm sorry, I'm not sure what the problem is here, could you please answer it? Or suggest a change

yShzZpp:
Dear Steve,
In this latest revision, I have diligently addressed the areas you highlighted and made the necessary modifications.
Additionally, I would like to extend my sincerest apologies for not paying attention to the formatting issues in my previous submissions. I have now taken the time to rectify all the formatting problems throughout the code.
Best regards,
zzpp

src/frustum_models/three_dimensional_lidar_frustum.cpp
```diff
@@ -101,7 +111,13 @@ bool ThreeDimensionalLidarFrustum::IsInside(const openvdb::Vec3d &pt)
 
   // // Check if inside frustum valid vFOV
   const double v_padded = fabs(transformed_pt[2]) + _vFOVPadding;
```
SteveMacenski, line 113: Please make sure to test, in detail, this logic, since we've found now 2 different errors in it that weren't caught before. test the boundaries of the different conditions and make sure this finally now works properly at the correct limit conditions

yShzZpp: I'm really sorry. I'm a beginner, and I've tried my best to locate where the problem might be, but with no success so far. I'm not sure if this will inconvenience you, but I still hope you could lend me a hand and point out where the issue lies. I'll do my best to fix it.

SteveMacenski: I don't know if there is an issue - I'm simply asking for you to write some tests to make sure this is correct in dealing with your different start/end FOV angles. If its not working, then look back at the geometry :-)

yShzZpp: I'm not sure how to write tests, but I can describe our application scenario to you. 
Our robot is a cuboid (1.00*0.5*0.3) with a front end that has a height of 0.7cm. Previously, we used 2D lidars, which only provided information about a single plane at a height of 0.2 meters. Now, due to new requirements, the robot needs to perceive obstacles below a height of 3 meters in front. Because the robot design is fixed, a 3D lidar can only be installed in the front part of the robot, at a height of 0.47 meters.

The lidar we are using is from [Livox](https://www.livoxtech.com/mid-360), with a FOV of -7° to 53°. Assuming our mounting angle is a roll angle of 90°, the angles it can see are illustrated in the diagram below.

Therefore, in the existing code, if I set the vertical FOV (vfov) to 114 degrees, points with positive values on the z-axis in the lidar's coordinate system will be correctly cleared. However, points with negative z-values or those below -7° will be cleared as well, which is not the desired outcome. On the other hand, if the vfov is set to a value less than 114 degrees, there will be residual points on the positive z-axis that cannot be cleared.
livox

SteveMacenski: I can appreciate that - so what's the problem? What aren't you able to test about your own code contribution to make sure it actually does what you want it to do? Previous iterations of this PR clearly didn't work properly so I can tell you didn't test in detail to make sure the behavior you want as working. What are you going to test now with the changes to ensure your changes work as expected?

PR #269, opened 2023-07-31
reset grid on disable

tonynajjar:
When disabling a source, it would make sense that its perceived voxels are cleared immediately as well and not left to expire

### DIFF
--- a/src/spatio_temporal_voxel_layer.cpp
+++ b/src/spatio_temporal_voxel_layer.cpp
@@ -457,6 +457,7 @@ void SpatioTemporalVoxelLayer::BufferEnablerCallback(
       response->message = "Enabling sensor";
     } else if (subcriber) {
       subcriber->unsubscribe();
+      ResetGrid();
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):

tonynajjar:
@SteveMacenski follow up question on this, `ResetGrid` will actually reset all the voxels from **all** the sensors right? Is there a way to reset only the voxels of the disabled source?

SteveMacenski:
No, the source is not stored in the voxel grid along with the time / marked status. It definitely **could** be, technically speaking, but that would add 8 bytes per voxel, effectively doubling the memory overhead of larger grids to store -- then require iterating through the grid to remove only those of a certain value. Disabling a sensor's data wouldn't be automatic

It is technically possible and wouldn't be overly difficult I don't think to add, but would be some effort to make that change

Another option which also had increased compute is that if you have 1 sensor you think you may need to disable regularly and want to have it cleared out, you could isolate that into its own instance of this layer.

Issue #270, opened 2023-08-01
Memory consumption increases over time


siferati:
## Description

When voxels are expired, they are set to inactive in the underlying grid by calling `openvdb::DoubleGrid::Accessor::setValueOff`

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/6077efe368780030eabd47835f103f256836b140/src/spatio_temporal_voxel_grid.cpp#L483

The problem with this is that the voxel still remains in the grid.
This means every voxel, even if inactive, will stay in memory for as long as STVL is running.

The only time when memory is freed is when the clear costmap service is requested and `openvdb::DoubleGrid::clear` is called.

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/6077efe368780030eabd47835f103f256836b140/src/spatio_temporal_voxel_grid.cpp#L426

Note that even the clear costmap recovery behavior doesn't free the memory, since that recovery behavior calls `costmap_2d::CostmapLayer::clearArea`, which also simply sets the voxels to inactive.

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/6077efe368780030eabd47835f103f256836b140/src/spatio_temporal_voxel_grid.cpp#L457

This means that under regular operation, where the clear costmap service isn't usually called, stvl memory consumption will keep increasing over time as the robot explores new areas of the map and sees new obstacles.

If you monitor the memory consumption with htop it's easy to see it increasing as the robot moves.

## Demo

Thankfully, there's a function to print the memory being used by the grid `openvdb::v6_2::DoubleGrid::memUsage`, so if you simply log that information the problem becomes obvious.

For example, add the following lines to the top of `SpatioTemporalVoxelGrid::ClearFrustums` function:

```cpp
setlocale(LC_NUMERIC, "");
ROS_ERROR("memory: %'lu", _grid->memUsage());
```

You can see how even though the obstacles (`stvl/voxel_grid` topic) are gone in rviz, the memory still keeps increasing over time.

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/assets/15384781/00b11a9d-d68e-43ca-ba9c-505d289494b7

## Solution

I looked through openvdb api and it doesn't seem like there is a way of removing a specific voxel after inserting it?
But we can achieve the desired results by calling `openvdb::Grid::prune`.

- I'm not sure if this is the best solution, but let's discuss in #271.

SteveMacenski:
[...]

PR #271, opened 2023-08-01
Fix memory consumption issue

siferati:
## Description

- Implementation of the solution proposed in #270.

By calling `openvdb::Grid::prune` at the end of `SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap` we can free the memory being taken by expired voxels.

I measured execution time of the prune function and it was under 1ms, so I don't think it should have any major impact on performance.

```cpp
const auto start = ros::WallTime::now();
_grid->pruneGrid();
ROS_ERROR_STREAM("dt: " << (ros::WallTime::now() - start).toNSec() / 1e6);
```

## Demo

After this fix, the memory taken by the grid goes down every time the voxels are expired.

However, note that htop doesn't reflect this and the memory consumption never goes down (although it's a lot lower than before).
I'm not sure why this is happening, but maybe something related with linux internal workings?
Even though the process isn't using the memory anymore, linux still thinks it is?
I'm not too familiar with this, but at least from the logs it's clear to see that the memory consumption of the grid does indeed go down every time voxels are expired.

You can also test this without the change in this PR:
- Let the robot move and memory consumption increase (let's say 10mb)
- Manually call clear costmap service, which will clear the grid
- Confirm that the logs do indeed show that the grid was cleared and memory was freed
- htop still shows stvl taking those 10mb of memory

### Before

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/assets/15384781/00b11a9d-d68e-43ca-ba9c-505d289494b7

### After

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/assets/15384781/cfc8e78e-6bc4-446a-9d22-4d8450c4526b

### DIFF
--- a/src/spatio_temporal_voxel_grid.cpp
+++ b/src/spatio_temporal_voxel_grid.cpp
@@ -250,6 +250,9 @@ void SpatioTemporalVoxelGrid::TemporalClearAndGenerateCostmap(                \
       PopulateCostmapAndPointcloud(pt_index);
     }
   }
+
+  // free memory taken by expired voxels
+  _grid->pruneGrid();
 }
 
 /*****************************************************************************/


SteveMacenski:
Seems reasonable to me! Can you just verify the complexity of the runtime of this sparsification process? I would just like to understand if this is a constant 1ms no matter the size of the grid or number of voxels expired being removed or if its proportionate to something that we may need to document if it gets really high in some situations

siferati:
@SteveMacenski I'm not too familiar with openvdb, but since they use a b-tree as the underlying structure I guess the complexity would scale logarithmically based on the amount of voxels in the tree?

I did some quick tests on a real robot.
Pruning the tree with ~100K voxels / ~1MB took less than 1ms.
Pruning with ~1M voxels / \~10MB took 2\~3ms.

SteveMacenski (approved):
Can you push this to `ros2` as well? I'll run bloom for ROS 1 and 2 with this change then

PR #288, opened 2024-04-25
Fix missing boost dependency

Timple:
As per: https://github.com/SteveMacenski/spatio_temporal_voxel_layer/issues/232#issuecomment-2078230991

### DIFF
--- a/openvdb_vendor/package.xml
+++ b/openvdb_vendor/package.xml
@@ -25,6 +25,7 @@
 
   <depend>libboost-iostreams-dev</depend>
   <depend>libboost-thread-dev</depend>
+  <depend>libgdal-dev</depend>  <!-- For libblosc-dev -->
   <depend>tbb</depend>
   <depend>zlib</depend>
 
--- a/openvdb_vendor/package.xml
+++ b/openvdb_vendor/package.xml
@@ -23,9 +23,9 @@
   <buildtool_depend>ament_cmake_vendor_package</buildtool_depend>
   <buildtool_depend>git</buildtool_depend>
 
+  <depend>libblosc-dev</depend>
   <depend>libboost-iostreams-dev</depend>
   <depend>libboost-thread-dev</depend>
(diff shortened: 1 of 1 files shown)


SteveMacenski:
Try cloning and removing STVL layer so just the vendor before rosdep & build. Make sure this isn't the only one

openvdb_vendor/package.xml
```diff
@@ -23,7 +23,9 @@
   <buildtool_depend>ament_cmake_vendor_package</buildtool_depend>
   <buildtool_depend>git</buildtool_depend>
 
+  <depend>libboost-iostreams-dev</depend>
   <depend>libboost-thread-dev</depend>
+  <depend>libgdal-dev</depend>  <!-- For libblosc-dev -->
```
Timple: `libblosc-dev` needs to be added to the rosdistro index for this to be listed properly. But it's way past bedtime here, so that's something for tomorrow :slightly_smiling_face: 

PR is in a working state now.

SteveMacenski:
Let me know when its ready for another release attempt!

Timple:
Pending: https://github.com/ros/rosdistro/pull/40865#pullrequestreview-2025777028

SteveMacenski:
Ah ok, let me know when that is all squared away :-)

Timple:
After this PR is merged, a new release can be triggered

SteveMacenski:
Release try 2 https://github.com/ros/rosdistro/pull/41067

Timple:
The build failed again. But is seems it failed just before the PR was merged. So I guess we'll have to wait another day..

Timple:
Ah, this fix is in the ros2 branch. But not on the humble branch.

How do you typically backport these changes?

SteveMacenski:
I don't have the same bots / processes setup here as I do in Nav2, so manually before cutting releases. This project doesn't get a huge number of PRs to need a great deal of automation. Usually I'm only backporting 2-3 merge commits

PR #290, opened 2024-05-12
Fix deps humble

Timple:
Backport because the humble builds still fail on this. 

Is there a reason there is a seperate `humble` branch which is identical to the `ros2` branch?

### DIFF
--- a/openvdb_vendor/package.xml
+++ b/openvdb_vendor/package.xml
@@ -23,6 +23,8 @@
   <buildtool_depend>ament_cmake_vendor_package</buildtool_depend>
   <buildtool_depend>git</buildtool_depend>
 
+  <depend>libblosc-dev</depend>
+  <depend>libboost-iostreams-dev</depend>
   <depend>libboost-thread-dev</depend>
   <depend>tbb</depend>
   <depend>zlib</depend>


SteveMacenski (approved):

SteveMacenski:
B/c we need to retain ABI/API compatibility with released distributions and `ros2` may diverge over time with new distributions.

Timple:
Advocate of the devil, why is there no iron branch :smile:

SteveMacenski:
Because I haven't released iron yet, humble is our current canary in the coal mine for the build farm :wink:

SteveMacenski:
https://github.com/ros/rosdistro/pull/41153 

Try 3

Timple:
Three times is ship-right!

Let's go for iron :sunglasses:

SteveMacenski:
Iron incoming

Timple:
https://build.ros2.org/job/Ibin_ujv8_uJv8__openvdb_vendor__ubuntu_jammy_arm64__binary/2/console
```
c++: fatal error: Killed signal terminated program cc1plus
```
This typically happens when stuff runs out of memory. I guess we can wait some more cycles? Otherwise we probably have to pass some flags to build less processes in parallel.

SteveMacenski:
https://build.ros2.org/job/Ibin_ujv8_uJv8__openvdb_vendor__ubuntu_jammy_arm64__binary/3/

It worked the 3rd time :-) 

I agree not ideal, but I suppose lets see if its happens on the regular before trying to do anything fancy. However if you knew what you wanted to try to reduce the build memory consumption, happy to merge + release so we can try it out. I just don't want us spinning our wheels if its 1/50 failure due to being on the edge of memory that is handled by auto-retries. If this was happening to us locally, I'd take that more seriously, but for the build farm, I don't really care as long as its not a 50%+ of the time issue. But if you feel separately, I'll never argue with someone wanting to improve things, just not something I'd prioritize myself :-)

Timple:
Fully agree!

I was hoping it would come to that. :)

PR #303, opened 2024-08-30
Fix arm64 build

Rayman:
The cmake module was located in another folder.

### DIFF
--- a/openvdb_vendor/CMakeLists.txt
+++ b/openvdb_vendor/CMakeLists.txt
@@ -5,6 +5,7 @@ find_package(ament_cmake REQUIRED)
 find_package(ament_cmake_vendor_package REQUIRED)
 
 list(APPEND CMAKE_MODULE_PATH "/usr/lib/x86_64-linux-gnu/cmake/OpenVDB")
+list(APPEND CMAKE_MODULE_PATH "/usr/lib/aarch64-linux-gnu/cmake/OpenVDB")
 find_package(OpenVDB 10 QUIET)
 
 ament_vendor(${PROJECT_NAME}


Timple:
That actually does make a lot of sense!

I was under the (wrong) impression that `openvdb` wasn't apt installable there...

SteveMacenski (approved):

SteveMacenski:
I'm backporting and releasing jazzy again

tfoote (commented):
This would be good to do conditionally or with a variable substitution so as not to pollute the path with unused elements.

Rayman:
@tfoote can you give me some pointers on which conditions this should be enabled.

Alternatively, what about searching for OpenVDBConfig.cmake in all paths under /usr and adding the parent folder?

tfoote:
In this case you should only be appending the path for the current architecture. Right now you're appending the path for both architectures, which probably will be ok, but could give some very weird errors if someone happens to be doing some multi-arch builds and it picks up the wrong architecture from their system. 

I don't have the exact commands but GNUInstallDirs is commonly used for picking the installation targets like this: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html#result-variables  so I presume it should be able to help set the search path. 

Otherwise you could simply check the arch in your CMake manually and then extend the CMAKE_MODULE_PATH conditionally.

SteveMacenski:
Good idea.

> Otherwise you could simply check the arch in your CMake manually and then extend the CMAKE_MODULE_PATH conditionally.

Do you have an example of best practices for this? I see ways of doing it with `CMAKE_SYSTEM_PROCESSOR` which seems a little fragile since its OS dependent, 32/64 bit dependent, and even Linux flavor dependent. If that's SOP, I'll do that, but otherwise I'd love to know if there's a better solution that some ROS core packages currently leverage

Rayman:
In the cmake these variables are available that could be of use:
```
-- CMAKE_CXX_LIBRARY_ARCHITECTURE=x86_64-linux-gnu
-- CMAKE_C_IMPLICIT_LINK_DIRECTORIES=/usr/lib/gcc/x86_64-linux-gnu/13;/usr/lib/x86_64-linux-gnu;/usr/lib;/lib/x86_64-linux-gnu;/lib
-- CMAKE_C_LIBRARY_ARCHITECTURE=x86_64-linux-gnu
-- CMAKE_LIBRARY_ARCHITECTURE=x86_64-linux-gnu
```

A solution could be to append `/usr/lib/${CMAKE_LIBRARY_ARCHITECTURE}/cmake/OpenVDB`?

PR #304, opened 2024-09-20
Relative Variants of Voxel and Passthrough Filters and Footprint Projection for Slope Navigation

JakubWarkocki:
#### **Summary**
This PR introduces new "relative" variants of the voxel and passthrough filters, which allow for more accurate obstacle detection on non-horizontal surfaces. By making the obstacle detection bounds relative to a customizable `z_reference_frame` (rather than the world frame), robots can now navigate uneven terrain without encountering the previous issues observed in Gazebo with the robot_localization and Nav2 stacks. The feature for clearing the robot footprint was also updated to provide an option to use 3D IMU orientation to project the footprint on the costmap.

These updates ensure improved behavior when broadcasting the robot's state with both 3D IMU orientation and `z` position on slopes, without causing unexpected navigation issues. The changes are implemented to ensure full backward compatibility with projects running the current version of SVTL.

#### **Problem Description**
- **Problems with using the robot's `z` position:** Since obstacle bounds are relative to the world frame, any change in the robot's `z` position altered the effective obstacle heights for the robot.
- **Incorrect Obstacle Detection on Slopes:** When using full 3D IMU orientation data, robots navigating slopes experienced improper obstacle detection, which could be dangerous, especially on downhill slopes. Obstacles' `z` coordinates in the world frame were not registered correctly since their global `z` coordinate was smaller than their height relative to the surface. On uphill slopes, the terrain itself could be mistakenly registered as an obstacle.
- **Inconsistent Localization on 2D Costmaps:** Using fully 2D localization created inconsistencies between GPS-based metrics and LiDAR-based metrics on a 2D costmap. This caused static obstacles to appear to "move" and become distorted as the robot moved on bumpy or sloped surfaces.

#### **Solution**
The newly introduced "relative" filter variants enable the robot’s obstacle detection to account for slopes by using the `z_reference_frame` parameter from the observation source, rather than relying on the world frame. This achieves:
- **Slope-Compatible Obstacle Detection:** The obstacle bounds adjust dynamically based on the robot’s current frame, allowing it to navigate traversable slopes and detect obstacles on them.
- **Enabling SVTL use while tracking the robot's `z` position:** This ensures consistent behavior regardless of altitude relative to the world frame.

> Note: With relative filters, obstacles directly in front of the robot may still be inside the footprint transformed without considering the robot's 3D rotation.

This issue was solved by introducing a footprint projection feature, which, if enabled, processes clearing the robot's footprint by transforming it using `tf2` instead of the `Costmap2D` footprint transformation. This provides an accurate projection of the footprint on the XY plane.

#### **Known Limitations**
- **Drastic Slope Changes:** The relative filters may still detect drastic slope changes as obstacles (e.g., transitioning from steep declines to inclines).
- **No Absolute Slope Limit Safeguard:** The current filters do not provide a mechanism to restrict obstacle detection based on a maximum slope. Further development with more complex point cloud analysis will be required to handle such cases.

#### **Usage**
To use the new features, the configuration file needs to be modified:
- Use the `"voxel_relative"` and `"passthrough_relative"` values for the filter attribute of the observation source.
- The `z_reference_frame` parameter, added to the observation source, sets the frame used to evaluate the `z` of each point of observation to determine if it is an obstacle. It defaults to the world frame of the layered costmap.
- `footprint_projection_enabled` activates projecting the footprint using `tf2` and is strongly recommended if using 3D localization. It defaults to `false`.
- `footprint_frame` sets the frame from which the footprint is transformed to the global frame. It is necessary to set this when using footprint projection. The recommended setting is the robot's base frame (manual setting is necessary because the parameter is inaccessible through the layered costmap).

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ Required dependencies ROS Kinetic, navigation, OpenVDB, TBB.
 
 An example fully-described configuration is shown below.
 
-Note: We supply two PCL filters within STVL to massage the data to lower compute overhead. STVL has an approximate voxel filter to make the data more sparse if very dense. It also has a passthrough filter to limit processing data within the valid minimum to maximum height bounds. The voxel filter is recommended if it lowers CPU overhead, otherwise, passthrough filter. No filter is also available if you pre-process your data or are not interested in performance optimizations. 
+Note: We supply two PCL filters within STVL to massage the data to lower compute overhead. STVL has an approximate voxel filter to make the data more sparse if very dense. It also has a passthrough filter to limit processing data within the valid minimum to maximum height bounds. The voxel filter is recommended if it lowers CPU overhead, otherwise, passthrough filter. No filter is also available if you pre-process your data or are not interested in performance optimizations. Filters have "relative" variants made to use robot's frame for obstacle classification instead of the global frame in case of navigating sloped surfaces.
 
 ```
 rgbd_obstacle_layer:
@@ -113,7 +113,9 @@ rgbd_obstacle_layer:
   observation_persistence: 0.0  #seconds
   max_obstacle_height:   2.0    #meters
   mark_threshold:        0      #voxel height
-  update_footprint_enabled: true
+  update_footprint_enabled: true 
+  footprint_projection_enabled: true #default off, recommended when using 3d IMU data. Uses tf2 to transform the footprint
+  robot_base_frame: "base_link" #frame to use with footprint_projection_enabled
   combination_method:    1      #1=max, 0=override
   obstacle_range:        3.0    #meters
   origin_z:              0.0    #meters
@@ -133,7 +135,8 @@ rgbd_obstacle_layer:
     observation_persistence: 0.0 #default 0, use all measurements taken during now-value, 0=latest 
     inf_is_valid: false          #default false, for laser scans
     clear_after_reading: true    #default false, clear the buffer after the layer gets readings from it
-    filter: "voxel"              #default passthrough, apply "voxel", "passthrough", or no filter to sensor data, recommended to have at one filter on
+    filter: "voxel"              #default passthrough, apply "voxel", "passthrough", "voxel_relative", "passthrough_relative", or no filter to sensor data, recommended to have at one filter on
+    z_reference_frame: "base_link" #default global_frame, use with *_relative filters as a frame for obstacle classification based on z coordinate
     voxel_min_points: 0          #default 0, minimum points per voxel for voxel filter
   rgbd1_clear:
     enabled: true                #default true, can be toggled on/off with associated service call
--- a/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
+++ b/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
@@ -180,7 +180,7 @@ private:
   rclcpp::Service<spatio_temporal_voxel_layer::srv::SaveGrid>::SharedPtr _grid_saver;
   std::unique_ptr<rclcpp::Duration> _map_save_duration;
   rclcpp::Time _last_map_save_time;
-  std::string _global_frame, _footprint_frame;
+  std::string _global_frame, _robot_base_frame;
   double _voxel_size, _voxel_decay;
   int _combination_method, _mark_threshold;
   volume_grid::GlobalDecayModel _decay_model;
@@ -191,7 +191,6 @@ private:
   boost::recursive_mutex _voxel_grid_lock;
 
   std::string _topics_string;
-
 
   // Dynamic parameters handler
   rclcpp::node_interfaces::OnSetParametersCallbackHandle::SharedPtr dyn_params_handler;
--- a/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
+++ b/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
@@ -76,7 +76,7 @@
 #include "message_filters/subscriber.h"
 #include "tf2_geometry_msgs/tf2_geometry_msgs.hpp"
 #include "tf2/buffer_core.h"
-
+#include "tf2_ros/buffer.h"
 namespace spatio_temporal_voxel_layer
 {
 
@@ -180,17 +180,18 @@ private:
   rclcpp::Service<spatio_temporal_voxel_layer::srv::SaveGrid>::SharedPtr _grid_saver;
   std::unique_ptr<rclcpp::Duration> _map_save_duration;
   rclcpp::Time _last_map_save_time;
-  std::string _global_frame;
+  std::string _global_frame, _footprint_frame;
   double _voxel_size, _voxel_decay;
   int _combination_method, _mark_threshold;
   volume_grid::GlobalDecayModel _decay_model;
-  bool _update_footprint_enabled, _enabled;
+  bool _update_footprint_enabled, _footprint_projection_enabled, _enabled;
   std::vector<geometry_msgs::msg::Point> _transformed_footprint;
   std::vector<observation::MeasurementReading> _static_observations;
   std::unique_ptr<volume_grid::SpatioTemporalVoxelGrid> _voxel_grid;
   boost::recursive_mutex _voxel_grid_lock;
 
   std::string _topics_string;
+
 
   // Dynamic parameters handler
   rclcpp::node_interfaces::OnSetParametersCallbackHandle::SharedPtr dyn_params_handler;
--- a/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
+++ b/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
@@ -135,6 +135,9 @@ public:
     std::shared_ptr<spatio_temporal_voxel_layer::srv::SaveGrid::Response> resp);
 
 private:
+  // clock
+  rclcpp::Clock::SharedPtr _clock;
+
   // Sensor callbacks
   void LaserScanCallback(
     sensor_msgs::msg::LaserScan::ConstSharedPtr message,
--- a/spatio_temporal_voxel_layer/src/measurement_buffer.cpp
+++ b/spatio_temporal_voxel_layer/src/measurement_buffer.cpp
@@ -54,7 +54,7 @@ MeasurementBuffer::MeasurementBuffer(
   const double & observation_keep_time, const double & expected_update_rate,
   const double & min_obstacle_height, const double & max_obstacle_height,
   const double & obstacle_range, tf2_ros::Buffer & tf, const std::string & global_frame,
-  const std::string & sensor_frame, const double & tf_tolerance,
+  const std::string & sensor_frame, const std::string & z_reference_frame, const double & tf_tolerance,
   const double & min_d, const double & max_d, const double & vFOV,
   const double & vFOVPadding, const double & hFOV,
   const double & decay_acceleration, const bool & marking,
@@ -67,7 +67,7 @@ MeasurementBuffer::MeasurementBuffer(
   _expected_update_rate(rclcpp::Duration::from_seconds(expected_update_rate)),
   _last_updated(clock->now()),
   _global_frame(global_frame), _sensor_frame(sensor_frame), _source_name(source_name),
-  _topic_name(topic_name), _min_obstacle_height(min_obstacle_height),
+  _topic_name(topic_name), _z_reference_frame(z_reference_frame), _min_obstacle_height(min_obstacle_height),
   _max_obstacle_height(max_obstacle_height), _obstacle_range(obstacle_range),
   _tf_tolerance(tf_tolerance), _min_z(min_d), _max_z(max_d),
   _vertical_fov(vFOV), _vertical_fov_padding(vFOVPadding),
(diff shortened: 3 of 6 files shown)


SteveMacenski:
It seems to me that you're filtering out the data below a minimum height when the robot is non-flat. Why not use then the `min_obstacle_height` parameter which does this as well in conjunction with setting your robot's base frame to your 6DOF frame to have the RPY enabled as part of the estimate? That would have the same effect without needing to change any software

JakubWarkocki:
I encountered the issues I described while using the `min_obstacle_height` and `max_obstacle_height` parameters in scenarios where the robot is moving on sloped surfaces or across multiple elevations. These parameters work well when the robot's movement is constrained to a flat surface, but issues arise in an operating environment in which robot's base Z-coordinate in the global frame changes with movement.

This happens because the `MeasurementBuffer` class transforms sensor data points from the `sensor_frame` to the`global_frame` and filters them based on their Z-coordinate in the global frame. While this works for a robot moving on a flat surface, it causes problems on slopes, stairs, or multi-floor environments, where the robot's base frame and the global frame (which remains fixed) can differ significantly in the Z-axis. This discrepancy results in obstacles being misclassified or ignored depending on whether their Z-coordinate lies within the defined bounds in the global frame.

For robots navigating uneven surfaces, it would be beneficial to temporarily switch to the robot's local frame to classify obstacles accurately, regardless of changes in elevation. This intermediate step would ensure the system remains robust, whether the robot is operating on slopes, or multiple flat surfaces, without relying solely on the global frame's Z-boundaries.

Another issue is clearing the robot's footprint. Default footprint transformation from `Costmap2d` only takes into account robot's yaw. On a sloped surface this causes the layer to clear obstacles immediately before the robot, as they are within the footprint that was transformed using only robot's yaw. This was solved by introducing an option to use `tf2` based footprint projection to clear the costmap.

SteveMacenski:
I spent some time to philosophically think about this. How I've thought about non-flat terrain was to segment out the ground via height, terrain estimation, AI, or other methods and then once the ground is roughly removed, to feed the remaining data into things like STVL or VL. The idea being that 'what is ground' and how to accurately remove it is very application dependent and its better to push that into a pre-stage than bake every one of `N` different methods that works on a different type of environment into the pre-processing stages of the costmap layers each. 

I don't think after reviewing this that this would effectively deal with remoing the ground for sloped navigation. I don't doubt that it makes it less-bad because you're working with the immediate base frame which removes the Z- related issues, but there can still be alot of local variation around the robot's base roll/pitch orientation (ex: you're on the side of a hill not pointed upward or downward of it, but sideways) that would mark ground that's navigable as obstacles 

I'm not against including this change however, but food for thought. I'm going to make some semi-pedantic code changes so we remove unnecessary TF lookups for when they're not required (and make the code more easily understandable for the fast observer).

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_layer.hpp
```diff
@@ -180,18 +180,19 @@ class SpatioTemporalVoxelLayer : public nav2_costmap_2d::CostmapLayer
...
-  bool _update_footprint_enabled, _enabled;
+  bool _update_footprint_enabled, _footprint_projection_enabled, _enabled;
   std::vector<geometry_msgs::msg::Point> _transformed_footprint;
   std::vector<observation::MeasurementReading> _static_observations;
   std::unique_ptr<volume_grid::SpatioTemporalVoxelGrid> _voxel_grid;
   boost::recursive_mutex _voxel_grid_lock;
 
   std::string _topics_string;
 
+
```
SteveMacenski: remove extra line

spatio_temporal_voxel_layer/src/measurement_buffer.cpp
```diff
@@ -137,44 +137,89 @@ void MeasurementBuffer::BufferROSCloud(
       return;
     }
```
SteveMacenski, line 139: For this block: move TF lookups into the `switch` statement so that we only lookup the few required. I don't want excess TF lookups for situations they're not used (TF lookups can add latency).

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
```diff
@@ -242,13 +251,20 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
     node->get_parameter(name_ + "." + source + "." + "model_type", model_type_int);
     ModelType model_type = static_cast<ModelType>(model_type_int);
 
+
```
SteveMacenski: Remove the extra line

spatio_temporal_voxel_layer/src/measurement_buffer.cpp
SteveMacenski: Plus if z reference frame is an empty string, that'll throw exceptions in TF if its attempted to look up

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
```diff
@@ -533,17 +549,43 @@ bool SpatioTemporalVoxelLayer::updateFootprint(
...
+        _transformed_footprint[i].x, _transformed_footprint[i].y,
+        min_x, min_y, max_x, max_y);
+    }
+  } else {
+    // Using tf2 for 3d rotation to provide accurate projection of the footprint
+    try {
+      geometry_msgs::msg::TransformStamped tf_footprint_stamped =
+        tf_->lookupTransform(
+          _global_frame, _footprint_frame,
+          rclcpp::Time(0));
```
SteveMacenski: Stamp this again the current time, not latest. Keeping sync is important

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
```diff
@@ -533,17 +549,43 @@ bool SpatioTemporalVoxelLayer::updateFootprint(
...
+          rclcpp::Time(0)); 
+      
+      _transformed_footprint = getFootprint();
+      for (unsigned int i = 0; i < _transformed_footprint.size(); i++) {
+        tf2::doTransform(_transformed_footprint[i], _transformed_footprint[i], tf_footprint_stamped);       
+        touch(
+          _transformed_footprint[i].x, _transformed_footprint[i].y,
+          min_x, min_y, max_x, max_y);
+      }
+
```
SteveMacenski: remove extra line

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
```diff
@@ -533,17 +549,43 @@ bool SpatioTemporalVoxelLayer::updateFootprint(
...
+    // Using tf2 for 3d rotation to provide accurate projection of the footprint
+    try {
+      geometry_msgs::msg::TransformStamped tf_footprint_stamped =
+        tf_->lookupTransform(
+          _global_frame, _footprint_frame,
+          rclcpp::Time(0)); 
+      
+      _transformed_footprint = getFootprint();
+      for (unsigned int i = 0; i < _transformed_footprint.size(); i++) {
+        tf2::doTransform(_transformed_footprint[i], _transformed_footprint[i], tf_footprint_stamped);
```
SteveMacenski, line 577: You're aware you're updating the stored member variable when you do this, right? Its not just updating it for local use in this method

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
```diff
@@ -108,6 +108,12 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
   // clear under robot footprint
   declareParameter("update_footprint_enabled", rclcpp::ParameterValue(true));
   node->get_parameter(name_ + ".update_footprint_enabled", _update_footprint_enabled);
+  // use 3d transform for clearing footprint
+  declareParameter("footprint_projection_enabled", rclcpp::ParameterValue(false));
+  node->get_parameter(name_ + ".footprint_projection_enabled", _footprint_projection_enabled);
+  // footprint frame ( used only with footprint projection enabled )
+  declareParameter("footprint_frame", rclcpp::ParameterValue(std::string("")));
```
SteveMacenski: Usually we call this "robot_base_frame" or "base_frame"

SteveMacenski (changes requested):
Add the new parameters / feature to the readme docs

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
JakubWarkocki: I am, original method also does this in order for the `_transformed_footprint` to be used in `updateCosts` method.

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
JakubWarkocki: I changed it but this causes issues with gazebo time so I still need to fix this

JakubWarkocki: lookup requires extrapolation into future if current time > latest time

README.md
```diff
@@ -113,7 +113,9 @@ rgbd_obstacle_layer:
   observation_persistence: 0.0  #seconds
   max_obstacle_height:   2.0    #meters
   mark_threshold:        0      #voxel height
-  update_footprint_enabled: true
+  update_footprint_enabled: true 
+  footprint_projection_enabled: true #default off, recommended when using 3d IMU data. Uses tf2 to transform the footprint
```
SteveMacenski, line 117: Set to `false` here, most people copy/paste this for use so it should be kept as a good default-centric baseline

SteveMacenski: I think somewhere in the readme, we need to have some explanation of your relative filters + how to set it up, so that its clear for later users how to use it and why its useful

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
```diff
@@ -78,7 +78,7 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
   RCLCPP_INFO(
     logger_, "%s's global frame is %s.",
     getName().c_str(), _global_frame.c_str());
-
+
```
SteveMacenski, line 81: Remove

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
```diff
@@ -108,6 +108,12 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
   // clear under robot footprint
   declareParameter("update_footprint_enabled", rclcpp::ParameterValue(true));
   node->get_parameter(name_ + ".update_footprint_enabled", _update_footprint_enabled);
+  // use 3d transform for clearing footprint
+  declareParameter("footprint_projection_enabled", rclcpp::ParameterValue(false));
+  node->get_parameter(name_ + ".footprint_projection_enabled", _footprint_projection_enabled);
+  // robot base frame ( necessary for 3d footprint projection )
+  declareParameter("robot_base_frame", rclcpp::ParameterValue(std::string("")));
```
SteveMacenski, line 115: Make it base_link or something reasonable please

SteveMacenski (changes requested):
Otherwise generally looks good now. I would like you to lint the file though, the formatting is inconsistent in places and its a bit distracting. Use `ament_cpplint` and `ament_uncrustify` and I can merge these in.

Please just verify to me that you tested all 4 modes and they continue to work as expected

spatio_temporal_voxel_layer/src/measurement_buffer.cpp
```diff
@@ -137,44 +137,98 @@ void MeasurementBuffer::BufferROSCloud(
...
+      break;
+    case Filters::VOXEL_RELATIVE :
+      tf_z_reference =
+        _buffer.lookupTransform(
+        _z_reference_frame, cloud.header.frame_id,
+        tf2_ros::fromMsg(cloud.header.stamp));
+      tf_global_from_z_reference =
+        _buffer.lookupTransform(
+        _global_frame, _z_reference_frame,
+        tf2_ros::fromMsg(cloud.header.stamp));
```
nilp-amin, line 213: Seems to be an unnecessary transform look up since the calculation in line 224 is redundant?

spatio_temporal_voxel_layer/src/measurement_buffer.cpp
```diff
@@ -137,44 +137,98 @@ void MeasurementBuffer::BufferROSCloud(
...
+      pcl_conversions::toPCL(*cld_transformed, *cloud_pcl);
+      sor.setInputCloud(cloud_pcl);
+      sor.setFilterFieldName("z");
+      sor.setFilterLimits(_min_obstacle_height, _max_obstacle_height);
+      sor.setDownsampleAllData(false);
+      sor.setLeafSize(v_s, v_s, v_s);
+      sor.setMinimumPointsNumberPerVoxel(static_cast<unsigned int>(_voxel_min_points));
+      sor.filter(*cloud_filtered);
+      pcl_conversions::fromPCL(*cloud_filtered, *cld_transformed);
+      tf2::doTransform(*cld_transformed, *cld_transformed, tf_global_from_z_reference);
```
nilp-amin, line 224: This transform calculation seems redundant.

nilp-amin (changes requested):
Was looking through this PR as this feature seems valuable and not much activity has happened on it for a while now.

nilp-amin:
Another question I had was is there a reason why an additional parameter `z_reference_frame` needs to be created when we already have the `robot_base_frame` parameter? Could we not just use this?

PR #306, opened 2024-11-05
Add arm64 version of OpenVDB to CMake module path.

### LINKED ISSUE #305: Arm64 Binaries for ROS 2 Jazzy
It seems that CI for ROS 2 Jazzy isn't working/enabled. The `ros-jazzy-spatio-temporal-voxel-layer` package is not found by apt on my arm64 system.

agoeckner:
This resolves #305.

As stated in https://github.com/SteveMacenski/spatio_temporal_voxel_layer/issues/305#issuecomment-2455928201, the CMake module path was not updated completely in #303 for arm64 builds.


SteveMacenski (approved):
Its a great day when we can guess a problem's solution quickly :-)

PR #308, opened 2024-12-03
Update in observation time comparison

### LINKED ISSUE #307: Inconsistent behavior for update rate between stvl and navigation2/nav2_costmap_2d
When comparing the behavior of the stvl and the nav2 obstacle layer some differences in behavior emerged during troubleshooting an issue. In the obstacle_layer (more precisely the used observation buffer) stale observations get purged when they are older than the specified `observation_keep_time_` (set via the `observation_persistence` parameter) compared to `clock_->now() [here](https://github.com/ros-navigation/navigation2/blob/dc26fa46d9240763af0cfc17f5b8a124f085eb41/nav2_costmap_2d/src/observation_buffer.cpp#L202). For the stvl, this is compared to the last update time [here](https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/0b76e8d84ca381bb4d72f4631bb30132d7c68d9c/spatio_temporal_voxel_layer/src/measurement_buffer.cpp#L221).

This leads to inconsistent behavior when sensor data is suddenly dropping out.
- For the obstacle_layer, the observations get removed from the buffer after the set time as would be expected. 
- The stvl keeps observations past the set `observation_persistence` as stale buffer entries get only removed when new observations are coming in (due to the update of `last_updated_`).

I suggest to replace the stale check in the stvl to compare with the current time instead of the `last_updated_`. Further, as commented in [this issue](https://github.com/ros-navigation/navigation2/issues/4734#issuecomment-2470214807), the update of `last_updated` in the reset function seems misplaced, and interferes with the `expected_update_rate` calculation.

Any thoughts on this?

sjusner:
The staleness of observations in the measurement buffer are now determined by comparison to clock_->now() instead of the time of last observation. Also, the last_updated is now not updated on reset.

Closes #307

### DIFF
--- a/spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
+++ b/spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
@@ -599,6 +599,11 @@ void SpatioTemporalVoxelLayer::reset(void)
 
   current_ = false;
   was_reset_ = true;
+
+  observation_buffers_iter it = _observation_buffers.begin();
+  for (; it != _observation_buffers.end(); ++it) {
+      (*it)->ResetLastUpdatedTime();
+  }
 }
 
 /*****************************************************************************/
--- a/spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
+++ b/spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
@@ -599,11 +599,6 @@ void SpatioTemporalVoxelLayer::reset(void)
 
   current_ = false;
   was_reset_ = true;
-
-  observation_buffers_iter it = _observation_buffers.begin();
-  for (; it != _observation_buffers.end(); ++it) {
-    (*it)->ResetLastUpdatedTime();
-  }
 }
 
 /*****************************************************************************/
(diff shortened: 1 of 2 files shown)


spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
```diff
@@ -599,11 +599,6 @@ void SpatioTemporalVoxelLayer::reset(void)
 
   current_ = false;
   was_reset_ = true;
-
-  observation_buffers_iter it = _observation_buffers.begin();
-  for (; it != _observation_buffers.end(); ++it) {
-    (*it)->ResetLastUpdatedTime();
```
SteveMacenski: This should remain, no? The costmap layers for the moment still do this https://github.com/ros-navigation/navigation2/blob/main/nav2_costmap_2d/plugins/obstacle_layer.cpp#L758-L774

SteveMacenski (changes requested):

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
sjusner: Yeah, that's true.
This change was a bit prematurely following [this issue](https://github.com/ros-navigation/navigation2/issues/4734). But that needs to be it's own PR.
Changes are now reverted on the reset() function.

SteveMacenski (approved):

PR #321, opened 2025-03-18
Fix ResetGridArea truth table logic

tonynajjar:
image
That's a XOR which is equivalent to `in_range != invert_area`.

Please double check, truth tables make me dizzy :dizzy_face: 

Btw I arrived to this by noticing that ClearCostmapExceptRegion and ClearCostmapAroundRobot were having reversed effects

If merged please also backport to Jazzy

### DIFF
--- a/spatio_temporal_voxel_layer/src/spatio_temporal_voxel_grid.cpp
+++ b/spatio_temporal_voxel_layer/src/spatio_temporal_voxel_grid.cpp
@@ -410,7 +410,7 @@ void SpatioTemporalVoxelGrid::ResetGridArea(
     const bool in_y_range = pose_world.y() > start.y && pose_world.y() < end.y;
     const bool in_range = in_x_range && in_y_range;
 
-    if(in_range == invert_area)
+    if(in_range != invert_area)
     {
       ClearGridPoint(pt_index);
     }


SteveMacenski:
I've not really used the clear around/except services myself in any applications, so I can see that it could be wrong. Though, this was added 3 years ago by @siferati because they needed it (and I entirely skipped it; since I didn't use it :wink:). 

Going through this for my own sanity mostly:
* `clearExceptRegionCallback` sets invert_area to `true` and `clearAroundRobotCallback` to `false`
* So `true` means that we should clear everything outside of the radius, but leave the radius alone
* In `SpatioTemporalVoxelGrid::ResetGridArea`, `in_range` means that that point is within the window.

So, with those definitions, I agree that your truth table makes sense, though _immediately suspicious_ given the chatGPT prompt, but I agree in this case it didn't give us garbage :laughing: .

I'm on board with this change and I imagine you tested this and you see the behavior you want from the proper service call, correct? If you validate that for me, happy to merge and backport (though I'll probably just change it in the UX rather than going through it locally, I don't have the backport bots setup on this repo)

tonynajjar:
Tested quickly today, will confirm more thoroughly tomorrow

tonynajjar:
Yup works as expected, ready to merge from my side

siferati:
Hi, thanks for tagging me @SteveMacenski !

_(Disclaimer: The explanation below is for ROS1 as I am not a user of ROS2)_

The code was mostly adapted from the default costmap implementation for `CostmapLayer::clearArea`.
You can see they do the same logic in [costmap_2d/src/costmap_layer.cpp#L21-L36](https://github.com/ros-planning/navigation/blob/4122c1b84296b7c49799ef1aee742565090753d9/costmap_2d/src/costmap_layer.cpp#L21-L36)
Their code looks a bit different, but it's essentially the same result:
```cpp
// ros default implementation
if(in_range != invert_area)
{
  continue;
}
ClearPoint();

// ---------

// my implementation in stvl
if(in_range == invert_area)
{
  ClearPoint();
}
```

Keep in mind that the main (and only) user of `CostmapLayer::clearArea` is the `clear_costmap_recovery` recovery behavior.
The `invert_area` arg was actually introduced for both the recovery behavior and the costmap layer at the same time:
- https://github.com/ros-planning/navigation/pull/1030

The default behavior is to clear the map **outside** of the given square.
And the inverted behavior is to clear **inside** the given square.
- `!invert` --> clear outside
- `invert` --> clear inside

You can confirm this is the intended behavior by the logs: [clear_costmap_recovery/src/clear_costmap_recovery.cpp#L100-L106](https://github.com/ros-planning/navigation/blob/4122c1b84296b7c49799ef1aee742565090753d9/clear_costmap_recovery/src/clear_costmap_recovery.cpp#L100-L106)
And it's also explained as such in its documentation: https://wiki.ros.org/clear_costmap_recovery
> \~<name>/invert_area_to_clear (bool, default: false)
> **Clears costmap/s outside of a square centered on the robot pose, if false and inside, if true.** The length of the side of the square is set by '~<name>/reset_distance' parameter.

## Summary

The current implementation in STVL follows the ROS1 implementation, and is correct when you consider it in the context of `clear_costmap_recovery` recovery behavior.

Unfortunately it seems like this has been changed for ROS2, so now the meaning has diverged when compared to ROS1:
- https://github.com/ros-navigation/navigation2/issues/2770

My suggestion for STVL would be to apply the fix in this PR only for ROS2 distros which have the fix linked above.
And keep the current code as is for everything else, including ROS1.
This would ensure the behavior of STVL stays consistent with that of ROS.

SteveMacenski:
Agreed! Thanks @siferati for the detailed comment and the links. I agree with your assessment and gives me better confidence about this change. Thanks!

ROS 2 only backports completed and linked below

PR #322, opened 2025-03-27
Added mapToWorldNoBounds Function

leander-dsouza:
> [!NOTE]  
> [#5010](https://github.com/ros-navigation/navigation2/pull/5010) must be merged prior.

## Description

* Added mapToWorldNoBounds function for fetching map coordinates as the  [mapToWorld](https://github.com/ros-navigation/navigation2/blob/main/nav2_costmap_2d/src/costmap_2d.cpp#L279) function only takes unsigned inputs.
* This fixes issues where, due to the unsigned type in the function arguments, the entire costmap would clear due to a **Bit Underflow** and set values such as `214747867`.

### DIFF
--- a/spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
+++ b/spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
@@ -961,8 +961,8 @@ void SpatioTemporalVoxelLayer::clearArea(
   // convert map coords to world coords
   volume_grid::occupany_cell start_world(0, 0);
   volume_grid::occupany_cell end_world(0, 0);
-  mapToWorld(start_x, start_y, start_world.x, start_world.y);
-  mapToWorld(end_x, end_y, end_world.x, end_world.y);
+  mapToWorldNoBounds(start_x, start_y, start_world.x, start_world.y);
+  mapToWorldNoBounds(end_x, end_y, end_world.x, end_world.y);
 
   boost::recursive_mutex::scoped_lock lock(_voxel_grid_lock);
   _voxel_grid->ResetGridArea(start_world, end_world, invert_area);


spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
```diff
@@ -961,8 +961,9 @@ void SpatioTemporalVoxelLayer::clearArea(
   // convert map coords to world coords
   volume_grid::occupany_cell start_world(0, 0);
   volume_grid::occupany_cell end_world(0, 0);
-  mapToWorld(start_x, start_y, start_world.x, start_world.y);
-  mapToWorld(end_x, end_y, end_world.x, end_world.y);
+
```
SteveMacenski: I think the inputs to `ResetGridArea` should not bound checked :smile: That way we clear everything that's in the data structure in that range, even when outside of the costmap's range (STVL can store a bit more than the occupancy grid)

The inputs to `CostmapLayer::clearArea` could be, but is redundant from your PR's change to that method.  So I don't think any bound checking in STVL is actually required

leander-dsouza: Yes, I agree the inputs to `ResetGridArea` should not be bound-checked, and I do not think it is currently happening in this change.
The issue is that [mapToWorld](https://github.com/ros-navigation/navigation2/blob/main/nav2_costmap_2d/src/costmap_2d.cpp#L279) expects unsigned coordinates, but there is no check for it.

SteveMacenski: I think we should use the `worldToMapNoBounds` version so that it can be negative

leander-dsouza: Thank you for the suggestion, Steve.
I have added a function titled `mapToWorldNoBounds` to get negative values as well.

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
```diff
@@ -961,14 +961,28 @@ void SpatioTemporalVoxelLayer::clearArea(
...
+  mapToWorldNoBounds(start_x, start_y, start_world.x, start_world.y);
+  mapToWorldNoBounds(end_x, end_y, end_world.x, end_world.y);
 
   boost::recursive_mutex::scoped_lock lock(_voxel_grid_lock);
   _voxel_grid->ResetGridArea(start_world, end_world, invert_area);
   CostmapLayer::clearArea(start_x, start_y, end_x, end_y, invert_area);
 }
 
+/*****************************************************************************/
+void SpatioTemporalVoxelLayer::mapToWorldNoBounds(
```
tonynajjar: isn't this the same as `Costmap2D::mapToWorld`?

leander-dsouza: The contents are the same; it is just that the map coordinates are `int` instead of  `unsigned int`.

tonynajjar: Ah okay, I think better to put it in Nav2 along with its other friends, other layers might want to use it as well

leander-dsouza: I have added this to the [nav2_costmap_2d](https://github.com/ros-navigation/navigation2/pull/5010) package, Tony.

tonynajjar (approved):

SteveMacenski (approved):

PR #324, opened 2025-04-01
Stvl distance decay

tonynajjar:
Implements distance-based decaying

Tested, works as intended

### DIFF
--- a/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
+++ b/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
@@ -130,7 +130,7 @@ public:
     rclcpp::Clock::SharedPtr clock,
     const float & voxel_size, const double & background_value,
     const int & decay_model, const double & voxel_decay,
-    const bool & pub_voxels, const double & voxel_distance_decay);
+    const double & voxel_distance_decay, const bool & pub_voxels);
   ~SpatioTemporalVoxelGrid(void);
 
   // Core making and clearing functions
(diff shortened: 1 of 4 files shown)


spatio_temporal_voxel_layer/src/spatio_temporal_voxel_grid.cpp
```diff
@@ -410,7 +432,7 @@ void SpatioTemporalVoxelGrid::ResetGridArea(
     const bool in_y_range = pose_world.y() > start.y && pose_world.y() < end.y;
     const bool in_range = in_x_range && in_y_range;
 
-    if(in_range == invert_area)
```
tonynajjar: I don't understand why this is a diff....

on the ros2 branch it's already != 

https://github.com/SteveMacenski/spatio_temporal_voxel_layer/blob/ros2/spatio_temporal_voxel_layer/src/spatio_temporal_voxel_grid.cpp#L413

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/spatio_temporal_voxel_grid.hpp
```diff
@@ -130,15 +130,16 @@ class SpatioTemporalVoxelGrid
     rclcpp::Clock::SharedPtr clock,
     const float & voxel_size, const double & background_value,
     const int & decay_model, const double & voxel_decay,
-    const bool & pub_voxels);
+    const bool & pub_voxels, const double & voxel_distance_decay);
```
SteveMacenski: Pedantic: but ordering this next to `voxel_decay` would be great to group like-things

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_grid.cpp
SteveMacenski: I'm guessing you just need to pull in main or rebase?

SteveMacenski (changes requested):
Minor nit, but otherwise LGTM

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_grid.cpp
tonynajjar: Not near my computer right now but this points to the current code and if you look at the current code on the ros2 branch, it's already `!=`

(It's late and most likely github is right and I'm wrong but I don't see it yet...)

SteveMacenski (approved):

tonynajjar:
Just checking, we can't backport to Jazzy because this breaks ABI right?

SteveMacenski:
Unfortunately so, I know some users that use the grid by itself so this would break them on a binary release :(

PR #341, opened 2025-10-27
Feature/support asymmetric lidar

pbrandschutz:
Added support of asymmetric LiDAR like the Livox MID360.

Instead of introducing 2 new params for
- vertical_fov_angle_top
- vertical_fov_angle_bottom
a parameter called "vertical_fov_offset" (default: 0.) is introduced and allows to set the offset angle between the XY-plane of the lidar TF and the center of the vertical FOV.
Using an additional param instead of 2 new params prevents existing implementations from becoming faulty.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ rgbd_obstacle_layer:
     min_z: 0.1                   #default 0, meters
     max_z: 7.0                   #default 10, meters
     vertical_fov_angle: 0.7      #default 0.7, radians
-    vertical_fov_offset: 0.3927  # default 0, radians. 3D Lidar only. Offset from planar axis. f.e. MID360 has a vFOV of -7deg to 52deg => offset +22.5deg (0.3927 rad)
+    vertical_fov_offset: 0.0     # default 0, radians. 3D Lidar only. Offset from planar axis. f.e. MID360 has a vFOV of -7deg to 52deg => offset +22.5deg (0.3927 rad)
     horizontal_fov_angle: 1.04   #default 1.04, radians
     decay_acceleration: 1.       #default 0, 1/s^2. If laser scanner MUST be 0
     model_type: 0                #default 0 (depth camera). Use 1 for 3D Lidar
--- a/README.md
+++ b/README.md
@@ -146,6 +146,7 @@ rgbd_obstacle_layer:
     min_z: 0.1                   #default 0, meters
     max_z: 7.0                   #default 10, meters
     vertical_fov_angle: 0.7      #default 0.7, radians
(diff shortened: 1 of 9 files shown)


spatio_temporal_voxel_layer/src/frustum_models/three_dimensional_lidar_frustum.cpp
```diff
@@ -84,16 +84,18 @@ bool ThreeDimensionalLidarFrustum::IsInside(const openvdb::Vec3d & pt)
   const double radial_distance_squared =
     (transformed_pt[0] * transformed_pt[0]) +
     (transformed_pt[1] * transformed_pt[1]);
-
+
```
SteveMacenski: Remove

spatio_temporal_voxel_layer/example/mid360.yaml
```diff
@@ -0,0 +1,46 @@
...
+        rgbd1_clear:
+          enabled: true                #default true, can be toggled on/off with associated service call
+          data_type: PointCloud2
+          topic: /livox/lidar
+          marking: false
+          clearing: true
+          max_z: 8.0                   # default 0, meters
+          min_z: 1.0                   # default 10, meters
+          vertical_fov_angle: 1.8151   # default 0.7, radians. For 3D lidars it's the symmetric FOV about the planar axis.
+          vertical_fov_offset: 0.3927  # default 0, radians. For 3D lidars it's the offset from the planar axis. f.e. a MID360 has a vFOV of -7deg to 52deg, so the offset is +22.5deg (0.3927 rad)
```
SteveMacenski: Please add this with a description to the readme parameter guide

SteveMacenski (approved):
Otherwise LGTM if you tested it! Please open it against `ros2` so its reflected in rolling and all new distributions

pbrandschutz:
@SteveMacenski it should be fine now.

README.md
```diff
@@ -146,6 +146,7 @@ rgbd_obstacle_layer:
     min_z: 0.1                   #default 0, meters
     max_z: 7.0                   #default 10, meters
     vertical_fov_angle: 0.7      #default 0.7, radians
+    vertical_fov_offset: 0.3927  # default 0, radians. 3D Lidar only. Offset from planar axis. f.e. MID360 has a vFOV of -7deg to 52deg => offset +22.5deg (0.3927 rad)
```
SteveMacenski: Set to `0.0` here so that if someone just copy+pastes they get a good default to start from (keep the `f.e.` comment though)

SteveMacenski (approved):
otherwise LGTM!

README.md
pbrandschutz: done

SteveMacenski:
Merged, thanks for the contribution!

PR #342, opened 2025-10-28
enable QoS policy overriding for publishers and subscribers

pashas567:
Add QosOverridingOptions::with_default_policies() to both publisher and
subscription options. This allows users to override QoS settings via
ROS 2 parameters (e.g., from a YAML file), enabling runtime tuning of
reliability, durability, and history depth without recompilation.

Useful for adapting to different network conditions or deployment
scenarios.


SteveMacenski (approved):
Please push for Rolling too so that the bleeding edge reflects changes for future distributions (maybe also Jazzy/Kilted too). I can merge them all as a set.

Sorry for the delay. ROSCon + vacation had me out of my desk for a couple of weeks.

PR #352, opened 2026-02-06
Feature/3D LiDAR frustrum visualization

ToRexZ:
## 3D Frustum Visualization

This feature adds **3D frustum visualization**, similar to the existing camera frustum visualization.

### Overview

The system generates **different frustums depending on the observation source behavior**:

- **Red frustum** for *clearing* observation sources  
- **Green frustum** for *marking* observation sources  

#### Clearing Frustum (LiVOX MID-360 LiDAR)

Clearing frustum visualization

#### Marking Frustum

Marking frustum visualization

### Cutoff Plane Inspection

In addition to the frustum itself, the visualization includes a **horizontal circle** with radius `max_z` at both:

- Minimum obstacle height  
- Maximum obstacle height  

This makes it easier to inspect the min and max cutoff planes used by the observation source.

### Visualization Controls

Each visual component is published under its own **marker namespace**, allowing individual elements to be toggled on or off as needed:

Marker namespace toggles

### Scaling and Parameters

The entire frustum scales dynamically based on the parameters provided to the corresponding observation source.

#### Supported Parameters

- Near and far clip (`min_z`, `max_z`)
- Vertical field of view (FOV)
- Vertical FOV offset
- Minimum and maximum obstacle height

---

Suggestions and feedback are very welcome 😄


spatio_temporal_voxel_layer/src/spatio_temporal_voxel_grid.cpp
```diff
@@ -374,25 +378,23 @@ void SpatioTemporalVoxelGrid::GetOccupancyPointCloud(
   sensor_msgs::PointCloud2Modifier modifier(*pc2);
 
   modifier.setPointCloud2Fields(
-    3,
-    "x", 1, sensor_msgs::msg::PointField::FLOAT32,
-    "y", 1, sensor_msgs::msg::PointField::FLOAT32,
+    3, "x", 1, sensor_msgs::msg::PointField::FLOAT32, "y", 1, sensor_msgs::msg::PointField::FLOAT32,
```
SteveMacenski: Undo this please

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_grid.cpp
```diff
@@ -281,14 +272,36 @@ void SpatioTemporalVoxelGrid::Mark(
...
+            frustum.SetVisualizationMode(true);
+            frustum.SetObstacleHeightBounds(
+              reading._min_obstacle_height, reading._max_obstacle_height);
+
+            frustum.TransformModel();
+          } else {
+          }
+        }
+      }
+#endif
```
SteveMacenski, line 293: Can we keep this isolated to the frustum models? I don't want to expose this into the grid class

spatio_temporal_voxel_layer/src/frustum_models/three_dimensional_lidar_frustum.cpp
```diff
@@ -65,12 +126,450 @@ ThreeDimensionalLidarFrustum::~ThreeDimensionalLidarFrustum(void)
...
+  }
+  if (_bottom_far_pts.empty()) {
+    return;
+  }
+
+  visualization_msgs::msg::MarkerArray msg_list;
+  visualization_msgs::msg::Marker marker;
+
+  // Common marker properties
+  const std::string frame_id("odom");  // Use costmap global_frame
```
SteveMacenski: This should not be hardcoded

SteveMacenski:
What are the 2 circles that are stacked & the yellow line? I'm unclear as to what those represent given the lidar seems to be at an angle in the first image.

I didn't review the vis marker population, that is *alot* of code. I think it needs to be vastly simplified or make use of some util functions / lambdas to reduce the boilerplate. 

I also think that the walls shouldn't be straight lines, its more 'donut' shaped which you can make using a `LINE_STRIP` for each line (which you could really just compute once and then revolve). 

The different marker namespaces seems a little overkill unless you can think of a good reason why someone would want 1-2 but not all (and all would be problematic). If it simplifies the code to use 1 namespace for each marking and clearing, I say that's a good trade off 

Overall, happy to have this and cool to see! Lets just polish the details a little more

ToRexZ:
Hi Steve.

The two circles represent the `min_obstacle_height` and `max_obstacle_height`, however i am open to other ways to visualize that, or even not include them at all. Perhaps two planes (cube with very small height) could visualize it better, or even constrain all the points used to create the entire frustum, along the `z-axis` in the `global_frame`. What do you think?

The yellow lines are not part of this PR, it is only the transform lines between `odom` <-> `base_link` <-> `sensor_frame` which i included in the pictures to get a sense of scale, and where the LiDAR is actually positioned. 

I will get to work on vastly simplifying the code. I completely agree. 

In terms of the shape, correct me if I am wrong, the way I understand the code in the `IsInside()` method, it is effectively checking the distance along the radial distance squared in the XY-plane from the center of the FOV, without considering the three dimensional euclidean distance to the point, hence the lines being straight and vertical represents the math better and not donut shaped. With that being said, I would assume that it would make great sense to have it consider three dimensional euclidean distance for spherical LiDARs, however for cylindrical layered 3D laser sensors, I believe it makes more sense to keep the radial distance. What do you think? Perhaps I can make a parameter that toggles that. 

Good idea with the namespaces, i will reduce it to one namespace pr marking/clearing. Do you think it would be a good idea to include the `observation_source` name in the namespace, in case there are different 3D LiDARs on the same robot?

ToRexZ:
After a more thorough review of the code, there appears to be an inconsistency in how the **marking** and **clearing** FOV are handled.

### Observations

- The **marking algorithm** does **not** explicitly consider the sensor FOV geometry.  
  Instead, it relies on:
  - `min_obstacle_height`
  - `max_obstacle_height`
  - a 3D Euclidean distance check from the sensor frame 
(This is perhaps the donut shape that you were refering to and led to a misunderstanding) 

- The **clearing algorithm**, on the other hand, uses a **radial distance–based FOV** with the **horizontal/vertical_fov_angle and offset**, meaning it applies a different spatial constraint than the marking step.

This mismatch may be intentional, but it introduces some non-trivial side effects.

### Resulting Issues

If the clearing FOV is **not tuned exactly** to match the LiDAR’s true FOV:

- **Clearing FOV smaller than the LiDAR FOV**  
  Marked voxels outside the clearing region will never be accelerated and will only decay passively. This leads to stale obstacles lingering longer than expected.

- **Clearing FOV larger than the LiDAR FOV**  
  Voxels that the sensor can never actually observe will receive accelerated clearing, which is physically inconsistent.

This behavior is especially noticeable when the robot is **stationary** and **dynamic obstacles** (e.g. people) move around it. In those cases, observations may be marked just outside the clearing FOV and persist longer than intended.

### Possible Improvement

One potential solution would be to apply the same **`IsInside` FOV logic** to both marking and clearing, in addition to the existing filters.  
This would ensure that voxels are **marked and cleared within an identical spatial region**, avoiding the tuning sensitivity described above.

**Potential drawback:**  
Using the same strict FOV boundaries for marking could cause some points near the edge of the sensor FOV to be excluded.

So:

Do you think this behavior should be:
- **Fixed in code** to unify the marking and clearing FOV logic, or
- **Explicitly documented** in the `README.md` as a known design choice and tuning consideration?

### Example

Below is an example where observations from a person walking around the robot are marked just *above* the clearing FOV, causing the issue described above. The higher the discrepancy, the more voxels are affected. 
image

_This is recorded with `voxel_decay=40.0` and `decay_acceleration=20.0` with `decay_model=0` (linear)_

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
```diff
@@ -241,7 +241,7 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
     model_type_int = node->declare_or_get_parameter(
       name_ + "." + source + "." + "model_type", 0);
     ModelType model_type = static_cast<ModelType>(model_type_int);
-
+
```
SteveMacenski: fix linting errors please - this PR introduces changes that creates many of them (not that it was perfect to begin with to be fair)

SteveMacenski:
> The two circles represent the min_obstacle_height and max_obstacle_height

Ah, OK. I suppose if the namespace makes this clear then all good :+1: 


> In terms of the shape, correct me if I am wrong, the way I understand the code in the IsInside() method, it is effectively checking the distance along the radial distance squared in the XY-plane from the center of the FOV, without considering the three dimensional euclidean distance to the point, hence the lines being straight and vertical represents the math better and not donut shaped. 

Upon re-reading, I agree. On https://github.com/SteveMacenski/spatio_temporal_voxel_layer/pull/352#issuecomment-3872315232, to be clear this is because clearing ranges check distance as XYZ but the marking is just XY? Then I agree that marking should check XYZ for range to include.

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/measurement_reading.h
```diff
@@ -41,17 +41,14 @@
...
 #include "geometry_msgs/msg/point.hpp"
 #include "geometry_msgs/msg/quaternion.hpp"
 #include "sensor_msgs/msg/point_cloud2.hpp"
 
-enum ModelType
-{
-  DEPTH_CAMERA = 0,
-  THREE_DIMENSIONAL_LIDAR = 1
-};
+enum ModelType { DEPTH_CAMERA = 0, THREE_DIMENSIONAL_LIDAR = 1 };
```
SteveMacenski: Another example; these are things that have no reason to be changed. Please revert unnecessary distracting changes please :-)

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/measurement_reading.h
```diff
@@ -62,19 +59,21 @@ struct MeasurementReading
   /*****************************************************************************/
   MeasurementReading()
   /*****************************************************************************/
-    : _cloud(std::make_shared < sensor_msgs::msg::PointCloud2 > ())
+  : _cloud(std::make_shared<sensor_msgs::msg::PointCloud2>())
```
SteveMacenski, line 66: Things like this are OK that actually fix formatting issues though

spatio_temporal_voxel_layer/src/frustum_models/depth_camera_frustum.cpp
```diff
@@ -99,41 +97,33 @@ void DepthCameraFrustum::ComputePlaneNormals(void)
...
   Eigen::Vector3d T_b(v_57.cross(v_45));
   T_b.normalize();
   _plane_normals.push_back(VectorWithPt3D(T_b[0], T_b[1], T_b[2], pt_[4]));
 
-  const Eigen::Vector3d v_67(pt_[7][0] - pt_[6][0],
-    pt_[7][1] - pt_[6][1], pt_[7][2] - pt_[6][2]);
-  const Eigen::Vector3d v_71(pt_[1][0] - pt_[7][0],
-    pt_[1][1] - pt_[7][1], pt_[1][2] - pt_[7][2]);
+  const Eigen::Vector3d v_67(pt_[7][0] - pt_[6][0], pt_[7][1] - pt_[6][1], pt_[7][2] - pt_[6][2]);
+  const Eigen::Vector3d v_71(pt_[1][0] - pt_[7][0], pt_[1][1] - pt_[7][1], pt_[1][2] - pt_[7][2]);
```
SteveMacenski: Any example: why is this changed in formatting? There's no reason to impact this as part of this PR :-)

ToRexZ:
Hey @SteveMacenski 

I would like your input on the marking frustum visualization.

As of right now, it is not possible to create a 3D Lidar frustum visualization that is guaranteed to match the actual marking algorithm. For example; the `horizontal_fov_angle` or `vertical_fov_angle` is not used by the `SpatioTemporalVoxelGrid::operator()` or in the `SpatioTemporalVoxelGrid::Mark()` method, meaning that it just assumes that if the sensor sees an object, it is marked. That is fine by itself, but it makes it impossible to create a 3D frustum visualization from the parameters: `obstacle_range_`, `min_obstacle_height` and `max_obstacle_height` alone. It is possible to create the frustums, but they will not actually match the algorithm. The only way is like a spherical segment, but it still needs to know the horizontal FOV. So i see two options: 

1. Get the corresponding frustum object for the observation source, apply that as a filter, making it parameterizable, and thus making the marking algorithm use the parameters. 
2. Remove the visualization of the marking observation, as I believe that it will create more confusion than clarity if it does not actually match the marking algorithm.

What method would you prefer? Once we have settled on a direction, I will get to work on implementing that strategy.

SteveMacenski:
I'm not sure I'm understanding clearly. Are you just saying that the visualizations don't include the min/max height (but represents the cylinder from the sensor FOV)? I think that's OK, the depth camera is just the camera frustum, not the 'modifiers' on top of it like min/max height that may cut through that frustum. Its for debugging the sensor model, not all the layer combination mechanics.

ToRexZ:
Sorry for the confusion :D I'm saying that the visualization frustum can only represent the clearing/acceleration algorithm. There is no easy way to visualize the marking algorithm for a 3D LiDAR since many of the parameters that are used to construct the frustum are actually not used in the marking algorithm. So i will either have to revert the changes of visualizing the marking frustum, or change the marking algorithm to check if the points are inside the frustum :)

SteveMacenski:
What's the concern? The marking should be range based which _is_ the FOV of the sensor that the clearing is just asking for the values of since we can't explicitly know that without configuration. Marking includes the FOV implicitly by... well being able to see things :laughing:

ToRexZ:
You are correct, the problem just becomes that it is impossible to construct the marking frustum, since it needs the explicit FOV parameters. The parameters on a marking observation source that are needed to construct the frustum does not actually represent the marking algorithm because it just marks whatever the sensor can see😉

The way I have created the marking frustum is the same way as the clearing frustum, by using `horizontal_fov_angle`, `vertical_fov_angle` and `vertical_fov_offset` and the range. However these parameters are not used in the marking algorithm, so changing these parameters on a marking observation source does not actually affect what it marks, it just affects the visualization of the frustum. That would confuse the user as they could create a very small marking frustum, but the sensor actually sees much wider. 

Another approach could be to dynamically compute the frustum by computing the min/max azimuth angle, and min/max altitude from the buffered cloud used for marking. That way the frustum actually represent what the sensor is actually capable of seeing, and not what the parameters are specifying😁

Hope this makes my concern a little more clear😉

SteveMacenski:
I think you should simply remove the marking part, lets not overcomplicate things :-)

PR #357, opened 2026-05-27
added obstruction polygons to 3d lidar frustrum

remco-r:
### TL;DR

I introduced the option to add "obstruction polygons" to the 3D lidar frustum. This way obstructions which are part of the robot body are considered out of field of view and persist accordingly.

lidar_fov_obstruction_stvl

_Green voxels are including obstruction polygons, white voxels are not (settings are slightly different in visualization so green and white do not correlate 1:1 in inside FOV region). It can be seen that in the old implementation the areas in obstructed FOV are cleared, while they persist in the new implementation._

Some documentation here may be outdated, check discussion below for changes and performance updates.

### Motivation

The concept that makes STVL fast is the assumption that everything inside the FOV is observed every update, therefore it can be cleared and filled with the latest observation without using expensive techniques such as raycasting. In my use case, the vehicle body often obstructed significant parts of the FOV of the 3D Lidar, invalidating the assumption. 
In my use case STVL was set up such that voxels outside of FOV persist for a few seconds. Sometimes there was an obstacle that got cleared when we drove past it and it was behind the obstruction (forklift mast), with the risk that our planner would now plan a path through the obstacle.

### Design

The user can add a list of obstruction polygons to the parameters. These polygons should be defined as a list of coordinates (azimuth elevation [rad] pairs) in lidar frame. When using the 3D Lidar frustum, these polygons will be loaded and when the frustums `isInside` method is called it will also check the obstruction polygons.

I implemented a ConvexPolygon2D class, that does precomputations on polygon normals to do efficient half plane comparisons with each point it has to process.

### Notes

* I tested using real world rosbags in jazzy. If desired I could make the PR to `ros2` branch as well.
* There were no tests in this package, i did not want to over complicate the PR by adding tests. There is a separate branch with tests https://github.com/remco-r/spatio_temporal_voxel_layer/tree/feature/obstruction_tests.
* Changes are not breaking. When no polygons are provided, the code should run as before. It also does not affect the depth camera frustum.
* I tried to make as little changes as possible to existing code and put most of the algo in a separate header.
* I have an AI assisted workflow, but checked every line of code and verified with unit tests.
* I vibe coded a python GUI to generate convex obstruction polygons from a pcd, also available in a separate branch https://github.com/remco-r/spatio_temporal_voxel_layer/tree/feature/obstruction_tests.

fov_boundary_tool_

### Changes

**Core feature `obstruction_polygons.hpp` (header-only)**
- `ConvexPolygon2D` struct with precomputed half-plane normals for O(n) point-in-polygon test
- Automatic winding order detection (CCW/CW) with correct inward normal computation
- `parsePolygonsFromString()` parses footprint-style bracket format: `"[[az1,el1, az2,el2, ...], [...]]"`
- `validatePolygons()` rejects degenerate, NaN, or out-of-range polygons with warnings
- `parseObstructionPolygonsFromParams()` declares and reads the ROS2 parameter

**Integration into STVL pipeline**
- `ThreeDimensionalLidarFrustum::IsInside()` after the existing hFOV check, computes azimuth/elevation and tests against obstruction polygons; points inside an obstruction polygon are not cleared
- `MeasurementBuffer` / `SpatioTemporalVoxelLayer` propagates polygons from parameter to frustum instances

**Configuration**
```yaml
observation_sources: source1
source1:
  # ... existing params ...
  obstruction_polygons: "[[az1,el1, az2,el2, az3,el3], [az4,el4, az5,el5, az6,el6]]"
```
Coordinates are in radians: azimuth [0, 2π], elevation [-π/2, π/2].

**GUI Tool `fov_boundary_tool` (Python/PyQt5) (outside of this PR)**
- Loads organized point clouds (ROS topic or PCD file), displays as range image
- User draws polygonal regions by clicking vertices on the range image
- Automatically decomposes concave selections into convex polygons
- Exports YAML and prints the STVL-compatible parameter string

### Limitations

- Polygons must be convex (the GUI tool handles decomposition)
- Does not support polygons crossing the 0/2π azimuth boundary (will check and reject)

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ An example fully-described configuration is shown below.
 
 Note: We supply two PCL filters within STVL to massage the data to lower compute overhead. STVL has an approximate voxel filter to make the data more sparse if very dense. It also has a passthrough filter to limit processing data within the valid minimum to maximum height bounds. The voxel filter is recommended if it lowers CPU overhead, otherwise, passthrough filter. No filter is also available if you pre-process your data or are not interested in performance optimizations. 
 
-```
+```yaml
 rgbd_obstacle_layer:
   enabled:               true
   voxel_decay:           20     #seconds if linear, e^n if exponential
@@ -150,22 +150,29 @@ rgbd_obstacle_layer:
 ```
 More configuration samples are included in the example folder, including a 3D lidar one.
 
+### Obstruction polygons for voxel retention
+
 In case you want to use the 3D lidar model with defined obstructions (e.g., robot body parts blocking the sensor's FOV), you can add the `obstruction_polygons` parameter. This prevents frustum clearing in the specified azimuth/elevation regions, preserving voxels behind blind spots:
 
 ```yaml
   lidar1_clear:
+    ...
     data_type: PointCloud2
     topic: /lidar/points
     marking: false
     clearing: true
     model_type: 1                #3D Lidar
-    horizontal_fov_angle: 6.28   #radians, full circle
-    vertical_fov_angle: 0.52     #radians
-    decay_acceleration: 5.0
+    horizontal_fov_angle: 6.28   #full circle
+    vertical_fov_angle: 1.57     #90 deg
+    obstacle_range: 20.0         #should be large enough to clear further than marking range, and further than voxels that may have been retained behind obstructions while robot moved.
     obstruction_polygons: "[[0.8,-0.2, 1.2,-0.2, 1.2,0.1, 0.8,0.1], [3.5,-0.3, 4.0,-0.3, 4.0,0.0, 3.5,0.0]]"
 ```
 
 Each inner bracket `[az1,el1, az2,el2, ...]` defines one convex polygon (minimum 3 vertices, azimuth in [0, 2π] radians, elevation in [-π/2, π/2] radians). If empty or unset, behavior is identical to the original STVL. Only applies to `model_type: 1`.
+
+<img width="340" height="314" alt="stvl_obstruction_polygon_retention" src="https://github.com/user-attachments/assets/e0828595-86f9-4935-b21d-989995b964fe" />
+
+_An example of 3d lidar with obstruction from forklift mast. The voxels in the obstructed FOV (orange) are not cleared, but retained for voxel_decay seconds_
 
 ### local/global_costmap_params.yaml
 
--- a/README.md
+++ b/README.md
@@ -150,7 +150,11 @@ rgbd_obstacle_layer:
 ```
 More configuration samples are included in the example folder, including a 3D lidar one.
 
-### Obstruction polygons for voxel retention
+### Obstruction polygons for voxel retention in blind spots
+
+<img width="340" height="314" alt="stvl_obstruction_polygon_retention" src="https://github.com/user-attachments/assets/e0828595-86f9-4935-b21d-989995b964fe" />
+
+_An example of 3d lidar with obstruction from forklift mast. The voxels in the obstructed FOV (orange) are not cleared, but retained for voxel_decay seconds_
 
 In case you want to use the 3D lidar model with defined obstructions (e.g., robot body parts blocking the sensor's FOV), you can add the `obstruction_polygons` parameter. This prevents frustum clearing in the specified azimuth/elevation regions, preserving voxels behind blind spots:
 
@@ -170,9 +174,13 @@ In case you want to use the 3D lidar model with defined obstructions (e.g., robo
 
 Each inner bracket `[az1,el1, az2,el2, ...]` defines one convex polygon (minimum 3 vertices, azimuth in [0, 2π] radians, elevation in [-π/2, π/2] radians). If empty or unset, behavior is identical to the original STVL. Only applies to `model_type: 1`.
 
-<img width="340" height="314" alt="stvl_obstruction_polygon_retention" src="https://github.com/user-attachments/assets/e0828595-86f9-4935-b21d-989995b964fe" />
+#### Edges are great-circle arcs, not az/el straight lines
 
-_An example of 3d lidar with obstruction from forklift mast. The voxels in the obstructed FOV (orange) are not cleared, but retained for voxel_decay seconds_
+The vertices will be represented as 3D *directions* from the sensor into the world. Each edge is the geodesic between two of them. This means in the polygon is **not** the flat box in (azimuth, elevation) that the numbers suggest. The edges will be bent according to  the following figure. Also, edges longer than pi will "flip" to cross the 0/2pi seam, because the arc is represented by the shortest distance between vertices.
+
+<img width="2247" height="957" alt="obstruction_polygon_projection" src="https://github.com/user-attachments/assets/32c2919d-0983-493d-91b7-dd8dd22a746a" />
+
+**Tip:** When you want to have a large azimuth polygon edge, split it into multiple smaller polygons to avoid bowing and flipping. Don't use multiple smaller azimuth edges in the same polygon, this will likely cause non convexity.
 
 ### local/global_costmap_params.yaml
 
--- a/README.md
+++ b/README.md
@@ -150,6 +150,23 @@ rgbd_obstacle_layer:
 ```
 More configuration samples are included in the example folder, including a 3D lidar one.
 
+In case you want to use the 3D lidar model with defined obstructions (e.g., robot body parts blocking the sensor's FOV), you can add the `obstruction_polygons` parameter. This prevents frustum clearing in the specified azimuth/elevation regions, preserving voxels behind blind spots:
+
+```yaml
+  lidar1_clear:
+    data_type: PointCloud2
+    topic: /lidar/points
+    marking: false
+    clearing: true
+    model_type: 1                #3D Lidar
+    horizontal_fov_angle: 6.28   #radians, full circle
+    vertical_fov_angle: 0.52     #radians
+    decay_acceleration: 5.0
+    obstruction_polygons: "[[0.8,-0.2, 1.2,-0.2, 1.2,0.1, 0.8,0.1], [3.5,-0.3, 4.0,-0.3, 4.0,0.0, 3.5,0.0]]"
+```
+
+Each inner bracket `[az1,el1, az2,el2, ...]` defines one convex polygon (minimum 3 vertices, azimuth in [0, 2π] radians, elevation in [-π/2, π/2] radians). If empty or unset, behavior is identical to the original STVL. Only applies to `model_type: 1`.
+
 ### local/global_costmap_params.yaml
 
 Add this plugin to your costmap params file. 
--- a/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
+++ b/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
@@ -57,7 +57,8 @@ class ThreeDimensionalLidarFrustum : public Frustum
 public:
   ThreeDimensionalLidarFrustum(
     const double & vFOV, const double & vFOVPadding,
-    const double & hFOV, const double & min_dist, const double & max_dist);
+    const double & hFOV, const double & min_dist, const double & max_dist,
+    std::shared_ptr<geometry::ObstructionFilter> obstruction_filter = nullptr);
   virtual ~ThreeDimensionalLidarFrustum(void);
 
   // Does nothing in 3D lidar model
@@ -69,9 +70,6 @@ public:
   // set pose of 3d lidar in global space
   virtual void SetPosition(const geometry_msgs::msg::Point & origin);
   virtual void SetOrientation(const geometry_msgs::msg::Quaternion & quat);
-
-  // Set obstruction polygons (blind spots in sensor-local angular space)
-  void SetObstructionFilter(std::shared_ptr<geometry::ObstructionFilter> filter);
 
 private:
   // utils to find useful frustum metadata
--- a/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
+++ b/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
@@ -84,9 +84,9 @@ private:
   // externally set postion and orientation of sensor in global space
   Eigen::Vector3d _position;
   Eigen::Quaterniond _orientation;
-  // computed transform matrices to apply to voxels during clearing
-  Eigen::Matrix3d _rotation;
-  Eigen::Vector3d _translation;
+  // inverse of the pose above, precomputed once to map voxels into sensor-local space
+  Eigen::Matrix3d _global_to_sensor_rotation;
+  Eigen::Vector3d _global_to_sensor_translation;
   bool _valid_frustum;
   bool _full_hFOV;
   std::shared_ptr<geometry::ObstructionFilter> _obstruction_filter;
--- a/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
+++ b/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
@@ -42,8 +42,11 @@
 
 // M_PI
 #include <cmath>
+#include <vector>
+#include <memory>
 // STVL
 #include "spatio_temporal_voxel_layer/frustum_models/frustum.hpp"
+#include "spatio_temporal_voxel_layer/obstruction_polygons.hpp"
 
 namespace geometry
 {
@@ -67,6 +70,10 @@ public:
   virtual void SetPosition(const geometry_msgs::msg::Point & origin);
   virtual void SetOrientation(const geometry_msgs::msg::Quaternion & quat);
 
+  // Set obstruction polygons (blind spots in sensor-local angular space)
+  void SetObstructionPolygons(
+    std::shared_ptr<std::vector<ConvexPolygon2D>> polygons);
+
 private:
   // utils to find useful frustum metadata
   double Dot(const VectorWithPt3D &, const openvdb::Vec3d &) const;
@@ -82,6 +89,7 @@ private:
   Eigen::Quaterniond _orientation_conjugate;
   bool _valid_frustum;
   bool _full_hFOV;
+  std::shared_ptr<std::vector<ConvexPolygon2D>> _obstruction_polygons;
 };
 
 }  // namespace geometry
--- a/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
+++ b/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
@@ -90,6 +90,7 @@ private:
   bool _valid_frustum;
   bool _full_hFOV;
   std::shared_ptr<std::vector<ConvexPolygon2D>> _obstruction_polygons;
+  ObstructionField _obstruction_field;
 };
 
 }  // namespace geometry
--- a/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
+++ b/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
@@ -71,8 +71,7 @@ public:
   virtual void SetOrientation(const geometry_msgs::msg::Quaternion & quat);
 
   // Set obstruction polygons (blind spots in sensor-local angular space)
-  void SetObstructionFilter(
-    std::shared_ptr<geometry::ObstructionFilter> filter);
+  void SetObstructionFilter(std::shared_ptr<geometry::ObstructionFilter> filter);
 
 private:
   // utils to find useful frustum metadata
@@ -84,9 +83,12 @@ private:
   double _min_d_squared, _max_d_squared;
   double _tan_vFOVhalf;
   double _tan_vFOVhalf_squared;
+  // externally set postion and orientation of sensor in global space
   Eigen::Vector3d _position;
   Eigen::Quaterniond _orientation;
-  Eigen::Quaterniond _orientation_conjugate;
+  // computed transform matrices to apply to voxels during clearing
+  Eigen::Matrix3d _rotation;
+  Eigen::Vector3d _translation;
   bool _valid_frustum;
   bool _full_hFOV;
   std::shared_ptr<geometry::ObstructionFilter> _obstruction_filter;
--- a/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
+++ b/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
@@ -71,8 +71,8 @@ public:
   virtual void SetOrientation(const geometry_msgs::msg::Quaternion & quat);
 
   // Set obstruction polygons (blind spots in sensor-local angular space)
-  void SetObstructionPolygons(
-    std::shared_ptr<std::vector<ConvexPolygon2D>> polygons);
+  void SetObstructionFilter(
+    std::shared_ptr<geometry::ObstructionFilter> filter);
 
 private:
   // utils to find useful frustum metadata
@@ -89,8 +89,7 @@ private:
   Eigen::Quaterniond _orientation_conjugate;
   bool _valid_frustum;
   bool _full_hFOV;
-  std::shared_ptr<std::vector<ConvexPolygon2D>> _obstruction_polygons;
-  ObstructionField _obstruction_field;
+  std::shared_ptr<geometry::ObstructionFilter> _obstruction_filter;
 };
 
 }  // namespace geometry
--- a/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
+++ b/spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
@@ -39,13 +39,13 @@
 #ifndef SPATIO_TEMPORAL_VOXEL_LAYER__OBSTRUCTION_POLYGONS_HPP_
 #define SPATIO_TEMPORAL_VOXEL_LAYER__OBSTRUCTION_POLYGONS_HPP_
 
-#include <vector>
+#include <algorithm>
 #include <cmath>
 #include <cstdint>
-#include <string>
 #include <sstream>
+#include <string>
 #include <utility>
-#include <algorithm>
+#include <vector>
 
 #include "rclcpp/rclcpp.hpp"
 
@@ -58,7 +58,10 @@ struct SphericalPoint
   double elevation;  // radians
 };
 
-struct Vec3D { double x, y, z; };
+struct Vec3D
+{
+  double x, y, z;
+};
 
 /**
  * @brief A convex polygon defined in (azimuth, elevation) space with precomputed
@@ -108,9 +111,8 @@ struct ConvexPolygon2D
       normals[i].z = dirs[i].x * dirs[j].y - dirs[i].y * dirs[j].x;
 
       // Check for degenerate edge (coincident vertices)
-      double len_sq = normals[i].x * normals[i].x +
-                      normals[i].y * normals[i].y +
-                      normals[i].z * normals[i].z;
+      double len_sq =
+        normals[i].x * normals[i].x + normals[i].y * normals[i].y + normals[i].z * normals[i].z;
 
       constexpr double kMinEdgeNormSquared = 1e-12;  // reject edges < ~0.001° apart
       if (len_sq < kMinEdgeNormSquared) {
@@ -127,9 +129,8 @@ struct ConvexPolygon2D
     }
 
     // flip normals if pointing outward (dot product with centroid < 0)
-    double test_dot = normals[0].x * centroid.x +
-                      normals[0].y * centroid.y +
-                      normals[0].z * centroid.z;
+    double test_dot =
+      normals[0].x * centroid.x + normals[0].y * centroid.y + normals[0].z * centroid.z;
     if (test_dot < 0.0) {
       for (auto & norm : normals) {
         norm.x = -norm.x;
@@ -143,9 +144,8 @@ struct ConvexPolygon2D
     constexpr double kConvexEps = 1e-9;  // tolerance for vertices lying on an edge's line
     for (size_t i = 0; i < n; ++i) {
       for (size_t k = 0; k < n; ++k) {
-        double side = normals[i].x * dirs[k].x +
-                      normals[i].y * dirs[k].y +
-                      normals[i].z * dirs[k].z;
+        double side =
+          normals[i].x * dirs[k].x + normals[i].y * dirs[k].y + normals[i].z * dirs[k].z;
         if (side < -kConvexEps) {
           return false;  // non-convex polygon: reject rather than mis-test
         }
@@ -178,8 +178,7 @@ struct ConvexPolygon2D
  * @return vector of valid, precomputed polygons
  */
 inline std::vector<ConvexPolygon2D> validatePolygons(
-  const std::vector<ConvexPolygon2D> & input,
-  const rclcpp::Logger & logger)
+  const std::vector<ConvexPolygon2D> & input, const rclcpp::Logger & logger)
 {
   std::vector<ConvexPolygon2D> valid;
   valid.reserve(input.size());
@@ -187,10 +186,7 @@ inline std::vector<ConvexPolygon2D> validatePolygons(
   for (size_t idx = 0; idx < input.size(); ++idx) {
     auto poly = input[idx];
     if (poly.vertices.size() < 3) {
-      RCLCPP_WARN(
-        logger,
-        "Obstruction polygon %zu has < 3 vertices, skipping.",
-        idx);
+      RCLCPP_WARN(logger, "Obstruction polygon %zu has < 3 vertices, skipping.", idx);
       continue;
     }
 
@@ -203,10 +199,7 @@ inline std::vector<ConvexPolygon2D> validatePolygons(
       }
     }
     if (has_nan) {
-      RCLCPP_WARN(
-        logger,
-        "Obstruction polygon %zu has NaN/inf vertices, skipping.",
-        idx);
+      RCLCPP_WARN(logger, "Obstruction polygon %zu has NaN/inf vertices, skipping.", idx);
       continue;
     }
 
@@ -230,8 +223,7 @@ inline std::vector<ConvexPolygon2D> validatePolygons(
     if (!poly.precompute()) {
       RCLCPP_ERROR(
         logger,
-        "Obstruction polygon %zu failed precomputation (non-convex or degenerate), skipping.",
-        idx);
+        "Obstruction polygon %zu failed precomputation (non-convex or degenerate), skipping.", idx);
       continue;
     }
 
@@ -246,8 +238,7 @@ inline std::vector<ConvexPolygon2D> validatePolygons(
  * Takes raw cartesian direction (x, y, z) from sensor frame.
  */
 inline bool isInsideAnyObstruction(
-  const std::vector<ConvexPolygon2D> & polygons,
-  double x, double y, double z)
+  const std::vector<ConvexPolygon2D> & polygons, double x, double y, double z)
 {
   for (const auto & poly : polygons) {
     if (poly.isInside(x, y, z)) {
@@ -276,7 +267,7 @@ struct ObstructionField
   };
   std::vector<Span> polygons;
 
-  bool empty() const {return polygons.empty();}
+  bool empty() const { return polygons.empty(); }
 };
 
 /**
@@ -297,8 +288,7 @@ inline double polygonAngularArea(const ConvexPolygon2D & p)
 /**
  * @brief Convert validated/precomputed polygons to flat SoA and sort by angular area
  */
-inline ObstructionField flattenAndSortPolygons(
-  const std::vector<ConvexPolygon2D> & polygons)
+inline ObstructionField flattenAndSortPolygons(const std::vector<ConvexPolygon2D> & polygons)
 {
   ObstructionField field;
 
@@ -342,15 +332,13 @@ inline ObstructionField flattenAndSortPolygons(
 /**
  * @brief Check if a 3D direction falls inside any obstruction polygon
  */
-inline bool isInsideAnyObstruction(
-  const ObstructionField & field,
-  float x, float y, float z)
+inline bool isInsideAnyObstruction(const ObstructionField & field, float x, float y, float z)
 {
-  const float* nx = field.nx.data();
-  const float* ny = field.ny.data();
-  const float* nz = field.nz.data();
+  const float * nx = field.nx.data();
+  const float * ny = field.ny.data();
+  const float * nz = field.nz.data();
 
-  for (const auto& span : field.polygons) {
+  for (const auto & span : field.polygons) {
     const uint32_t end = span.start + span.count;
     int any_negative = 0;
     for (uint32_t i = span.start; i < end; ++i) {
@@ -373,17 +361,16 @@ inline bool isInsideAnyObstruction(
  * @return vector of parsed (unvalidated) polygons
  */
 inline std::vector<ConvexPolygon2D> parsePolygonsFromString(
-  const std::string & input,
-  const rclcpp::Logger & logger)
+  const std::string & input, const rclcpp::Logger & logger)
 {
   std::vector<ConvexPolygon2D> polygons;
 
   // Find the outer brackets
   size_t outer_start = input.find('[');
   size_t outer_end = input.rfind(']');
-  if (outer_start == std::string::npos || outer_end == std::string::npos ||
-    outer_end <= outer_start)
-  {
+  if (
+    outer_start == std::string::npos || outer_end == std::string::npos ||
+    outer_end <= outer_start) {
     RCLCPP_WARN(logger, "Obstruction polygons string has invalid format.");
     return polygons;
   }
@@ -411,13 +398,14 @@ inline std::vector<ConvexPolygon2D> parsePolygonsFromString(
     while (std::getline(ss, token, ',')) {
       // Trim whitespace
       size_t start = token.find_first_not_of(" \t");
-      if (start == std::string::npos) {continue;}
+      if (start == std::string::npos) {
+        continue;
+      }
       try {
         values.push_back(std::stod(token.substr(start)));
       } catch (const std::exception &) {
         RCLCPP_WARN(
-          logger, "Obstruction polygon %d: failed to parse value '%s'.",
-          poly_idx, token.c_str());
+          logger, "Obstruction polygon %d: failed to parse value '%s'.", poly_idx, token.c_str());
       }
     }
 
@@ -429,8 +417,7 @@ inline std::vector<ConvexPolygon2D> parsePolygonsFromString(
       polygons.push_back(std::move(poly));
     } else if (!values.empty()) {
       RCLCPP_WARN(
-        logger,
-        "Obstruction polygon %d needs >= 6 values (3 x/y pairs), got %zu. Skipping.",
+        logger, "Obstruction polygon %d needs >= 6 values (3 x/y pairs), got %zu. Skipping.",
         poly_idx, values.size());
     }
 
@@ -449,11 +436,9 @@ inline std::vector<ConvexPolygon2D> parsePolygonsFromString(
  *
  * For STVL use, x=azimuth (rad), y=elevation (rad).
  */
-template<typename NodeT>
+template <typename NodeT>
 std::vector<ConvexPolygon2D> parseObstructionPolygonsFromParams(
-  NodeT node,
-  const std::string & param_prefix,
-  const rclcpp::Logger & logger)
+  NodeT node, const std::string & param_prefix, const rclcpp::Logger & logger)
 {
   std::string param_name = param_prefix + ".obstruction_polygons";
   std::string polygons_str;
@@ -470,9 +455,7 @@ std::vector<ConvexPolygon2D> parseObstructionPolygonsFromParams(
 
   if (!polygons.empty()) {
     RCLCPP_INFO(
-      logger,
-      "Parsed %zu obstruction polygon(s) for %s",
-      polygons.size(), param_prefix.c_str());
+      logger, "Parsed %zu obstruction polygon(s) for %s", polygons.size(), param_prefix.c_str());
   }
 
   return validatePolygons(polygons, logger);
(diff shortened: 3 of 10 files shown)


SteveMacenski:
This is neat, thanks! 

A question: why do we see alot of green "noise" voxels that aren't in your clear obstruction polygon along the wall? It seems like some things are being miscalculated. I think some answers to this are necessary before I give the code a detailed review. 

We would want this directed towards `main` too after this is approved so that its reflected in Lyrical and newer.

remco-r:
Thanks for your response @SteveMacenski! 

The "green noise" is a configuration choice. When voxels are outside the FOV they persist, when they come inside the FOV again later, they are only cleared up to the clearing range (`max_z`). In the gif above, the same sensor source is used for marking and clearing, both with `max_z = 13` meters. So the noise you see are voxels outside the clearing range, which will then decay based on the time constraint.

To demonstrate this I did another run with separate marking and clearing sources (but the same actual sensor source), and setting `max_z = 100` for the clearing. I also added a grey disc indicating the marking range. Now you can see all voxels that come back into the FOV being cleared immediately.

Thank you for having a look! :smile: 

fov-2026-06-05_18 59 04

SteveMacenski:
That gif is really nice - maybe worth documenting that configuration, I'm sure it'll be useful to some! 

I'm not 100% understanding the phenomenon with green "noise" voxels on otherwise flat walls that are fully in the FOV of the sensor. I thought that was only showing voxels currently detected as in the obstruction area, but they are clearly not. There's clearly live data that's overlapping with it. I'm referring to the green voxels that overlap with the grey (not the trailing green ones or the blocks of green ones in the obstruction FOV). Unless I'm mistaken, that seems to indicate to me a bug in how the FOV is being computed if those cells are being mis-classified as such. 

Can you clarity?

In general though beyond that point, I love this idea and very supportive of adding it :-)

remco-r:
Thank you.

I think i now understand what you mean. In the gif in the PR description, there are a couple of "random" green voxels popping up between the white ones. Some encircled a in red below.

image

The green voxels in the PR descriptions' gif are not only visualizing outside FOV voxels. It's actually displaying a rosbag from a few months ago, where STVL voxels where already present (white), and in parallel I'm running the new STVL, displaying **all** voxels in green. I made the white voxels slightly bigger so that they should overshadow the green ones.

Why do you see the noise? Before I believed it to be slightly different SVTL configurations, but even when i run two identical configurations parallel to each other i get the noise. I believe its slight timing differences together with a slightly drifting odom frame. The two instances dont have time synchronization and do tf lookups for the same cloud at different times. Some points that are just on the edge of a voxel discretization will show in green and not in white. I cross checked and there are also a few white voxels that don't show up in green. 

I made another gif running two identical STVL configurations except one has obstruction polygons. Now displayed in real time.

fov-2026-06-08_18 37 57

If you feel this is an issue, i can look into time synchronization and the lookup differences.

remco-r:
I was curious about the performance. I benchmarked `ThreeDimensionalLidarFrustum::IsInside` and found that adding a single obstruction polygon increases its cost by ~2.6x, while adding further polygons only scales by ~4%   each. This is because the dominant cost is the atan2/sqrt conversion from cartesian to azimuth/elevation, a one-time computation per point that's needed as soon as any polygon is present. The actual half-plane polygon test is cheap in comparison.

There are ways to optimize this, but profiling the full costmap executable with perf shows that IsInside accounts for only ~5% of total CPU time (with polygons enabled). I dont think the added complexity will be justified for that margin. However, the analysis gives a good idea of the performance of the obstruction polygons addition.

<details>

<summary> Benchmark results </summary>


Benchmarked with state variables `num_points` (amount of voxels to check) and `num_polygons`

Used typical 3d lidar setup, like in the gifs. Each polygon takes 10% of the FOV (full vertical, 10% horizontal). 30% of the points outside of FOV.

```
Run on (12 X 5000 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 256 KiB (x6)
  L3 Unified 12288 KiB (x1)
Load Average: 0.79, 1.62, 1.72
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
-----------------------------------------------------------------------------------------------------
Benchmark                                           Time             CPU   Iterations UserCounters...
-----------------------------------------------------------------------------------------------------
IsInsideBenchmark/IsInsideBatch/0/1000/0         19.6 us         19.6 us        35693 items_per_second=51.1267M/s points=1k polygons=0
IsInsideBenchmark/IsInsideBatch/0/10000/0         212 us          212 us         3428 items_per_second=47.1962M/s points=10k polygons=0
IsInsideBenchmark/IsInsideBatch/0/100000/0       2101 us         2101 us          336 items_per_second=47.5942M/s points=100k polygons=0
IsInsideBenchmark/IsInsideBatch/1/1000/0         50.8 us         50.8 us        10000 items_per_second=19.6916M/s points=1k polygons=1
IsInsideBenchmark/IsInsideBatch/1/10000/0         551 us          551 us         1262 items_per_second=18.1518M/s points=10k polygons=1
IsInsideBenchmark/IsInsideBatch/1/100000/0       5505 us         5505 us          130 items_per_second=18.1659M/s points=100k polygons=1
IsInsideBenchmark/IsInsideBatch/2/1000/0         53.6 us         53.6 us        13022 items_per_second=18.6575M/s points=1k polygons=2
IsInsideBenchmark/IsInsideBatch/2/10000/0         575 us          575 us         1224 items_per_second=17.3848M/s points=10k polygons=2
IsInsideBenchmark/IsInsideBatch/2/100000/0       5763 us         5762 us          109 items_per_second=17.3548M/s points=100k polygons=2
IsInsideBenchmark/IsInsideBatch/3/1000/0         57.2 us         57.2 us        12563 items_per_second=17.4735M/s points=1k polygons=3
IsInsideBenchmark/IsInsideBatch/3/10000/0         583 us          583 us         1213 items_per_second=17.155M/s points=10k polygons=3
IsInsideBenchmark/IsInsideBatch/3/100000/0       5819 us         5818 us          117 items_per_second=17.1881M/s points=100k polygons=3
```
</details>



flamegraph_perf

_isInside part of full costmap executable perf while running on same rosbag as in the gif_

SteveMacenski:
Do you still see a few voxels that are wrong? Maybe its worth running STVL with your change and coloring the voxels or something within a single instance running so we can see the actual total state of the voxel grid to validate the results. I understand why you did what you did, but it doesn't give me confidence it fully works given we still see that noise.

I think you can hack into this 2 pointclouds: 1 that is the in, and other that is out of frustum - as a temporary validation measure. I'd want to see the internal state of the layer as correct, not seeing a diff(ish) from 2 instances. 

> found that adding a single obstruction polygon increases its cost by ~2.6x

That's pretty significant. Maybe it is worth optimizing that a bit. I am pretty performance sensitive about this layer. You're also using it in the 'simpliest' mode possible (1x 3D lidar) whereas many use it with N RGBD cameras that each produce multiples of the amount of data that the 3D lidar provides.

spatio_temporal_voxel_layer/src/frustum_models/three_dimensional_lidar_frustum.cpp
```diff
@@ -112,6 +114,21 @@ bool ThreeDimensionalLidarFrustum::IsInside(const openvdb::Vec3d & pt)
...
 
+  // Check if the point falls inside any obstruction polygon (blind spot)
+  // Regard points in blind spots as outside the frustum (i.e., not cleared)
+  if (_obstruction_polygons && !_obstruction_polygons->empty()) {
+    double azimuth = std::atan2(transformed_pt[1], transformed_pt[0]);
+    if (azimuth < 0.0) {
+      azimuth += 2.0 * M_PI;  // shift to [0, 2pi] to match polygon convention
+    }
+    const double xy_dist = std::sqrt(radial_distance_squared);
+    const double elevation = std::atan2(transformed_pt[2], xy_dist);
```
SteveMacenski: Yeah these look like pretty rough computations.

remco-r:
**1. New gif**

Made new gif with your suggestions. This is a single instance publishing also "outside of frustum" voxels.

stvl_obstruction-2026-07-04_16 17 17

**2. Optimization of isInside()**

`isInside()` now only does dot product of precomputed polygon half-circle planes and rays from sensor frame to voxels. This way the conversion from Euclidean to spherical coordinates is avoided at runtime, at the cost of precomputation on startup.

half_plane_visualization

_Previous implementation on left, new 3D implementation on right. Shows tests for an inside (green) and outside (red) point for one of the polygons edges._

**3. Performance**

* Profiling now shows about 1.5% of full costmap executable is `isInside()` vs about 5% previously. (these measurements are very sensitive to environment, so should be taken it with a grain of salt.)
* Benchmark of 3D isInside now shows much lower hot path computation time increase for `isInside` when polygons are present, from ~2.6x before to ~1.2x now (0→3 polygons, showing +28% first polygon, +12% each additional)

<details>

<summary> Benchmark results </summary>


Benchmarked with state variables `num_points` (amount of voxels to check) and `num_polygons`

Used typical 3d lidar setup, like in the gifs. Each polygon takes 10% of the FOV (full vertical, 10% horizontal). 30% of the points outside of FOV.

```
Run on (12 X 5000 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x6)
  L1 Instruction 32 KiB (x6)
  L2 Unified 256 KiB (x6)
  L3 Unified 12288 KiB (x1)
Load Average: 0.48, 0.60, 0.79
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
-----------------------------------------------------------------------------------------------------
Benchmark                                           Time             CPU   Iterations UserCounters...
-----------------------------------------------------------------------------------------------------
IsInsideBenchmark/IsInsideBatch/0/1000/0         19.7 us         19.7 us        35144 items_per_second=50.8363M/s points=1k polygons=0
IsInsideBenchmark/IsInsideBatch/0/10000/0         209 us          209 us         3372 items_per_second=47.8595M/s points=10k polygons=0
IsInsideBenchmark/IsInsideBatch/0/100000/0       2137 us         2135 us          329 items_per_second=46.8433M/s points=100k polygons=0
IsInsideBenchmark/IsInsideBatch/1/1000/0         21.2 us         21.1 us        32385 items_per_second=47.3127M/s points=1k polygons=1
IsInsideBenchmark/IsInsideBatch/1/10000/0         276 us          276 us         2528 items_per_second=36.2428M/s points=10k polygons=1
IsInsideBenchmark/IsInsideBatch/1/100000/0       2779 us         2778 us          249 items_per_second=35.9941M/s points=100k polygons=1
IsInsideBenchmark/IsInsideBatch/2/1000/0         24.0 us         24.0 us        28686 items_per_second=41.6831M/s points=1k polygons=2
IsInsideBenchmark/IsInsideBatch/2/10000/0         312 us          312 us         2224 items_per_second=32.0828M/s points=10k polygons=2
IsInsideBenchmark/IsInsideBatch/2/100000/0       3103 us         3102 us          226 items_per_second=32.2362M/s points=100k polygons=2
IsInsideBenchmark/IsInsideBatch/3/1000/0         26.7 us         26.7 us        26774 items_per_second=37.4989M/s points=1k polygons=3
IsInsideBenchmark/IsInsideBatch/3/10000/0         354 us          354 us         1981 items_per_second=28.2336M/s points=10k polygons=3
IsInsideBenchmark/IsInsideBatch/3/100000/0       3523 us         3522 us          200 items_per_second=28.3952M/s points=100k polygons=3

```
</details>

SteveMacenski:
OK great! I :+1: the math now from that visualization 

For the perf numbers: any way we can improve that further? Admittedly 28% + 12% is very non-trivial considering everything that's happening in STVL by itself. This is a big improvement in the right direction though. I would, at least naively as to intuition, think we could do something like this for half that cost. 

It could be as simple as inlining some of this to keep information locality:

```
  // Check if the point falls inside any obstruction polygon (blind spot)
  if (_obstruction_polygons && !_obstruction_polygons->empty()) {
    if (geometry::isInsideAnyObstruction(*_obstruction_polygons,
        transformed_pt[0], transformed_pt[1], transformed_pt[2])) {
      return false;  // point is in a blind spot
    }
  }
```

To doing the for loop right here including the math (don't call any methods --> do it all here). I've often noticed in high-performance computing that `for (...) callMethod()` is much slower than `for (...) //do method`.

remco-r:
I share your suspicion. The call stack is `isInsideAnyObstruction` -> for each polygon -> `isInside` -> dot products. This should scale linearly with amount of points if evenly distributed, not have an extra increase for the first polygon. The only thing i can come up with is cpu process flow (which is not my expertise). From 1 polygon there is a cpu branch that needs to be added, which is a fixed overhead.

I tried several optimizations, with little success:
* Your suggestion to have the function inline in `isInsideAnyObstructions` -> no significant difference
* Invert the loop to: for each polygon { for each point } -> worse
* Different data structures to store polygon normals (like flat hierarchy) -> no significant difference
* Reduce precision of points to mm and do integer product -> no significant difference

The only thing that made a difference was the order of checking polygon edges. This can reduce the amount of checks per point using early return for points outside the polygon. However, the improvements are highly dependent on polygon size, shape and point distribution. 
The following image shows an extreme use case where such an optimization will pay off. Testing in the indicated order was ~40% faster than sequential order. ~80% of the points only need the check with the first edge to already fail.

sorting_visualization_full_fov

I can implement some optimal form of normal reordering. I do want to mention `isInside` accounts for only 1.5% of the full costmap executable (in my 3D lidar config, given the previous 3D normal optimization). Here the savings would be only ~0.6% overall in the best case. Also, obstruction polygons are now only applicable to the 3D lidar frustum, not to the depth camera frustum.

SteveMacenski:
I'm less concerned with the total % on your CPU for the costmap process and more relative for this layer since that's what this is changing (and I know from experience that many sensors and long range configurations will have varying performance). If we hit a wall on perf, then we can stop there and I can review the code for merging. However while we're still making improvements, I'd like to see if we can make such an improvement for another 1x if possible. 

A question though: do you have an RGBD camera handy? If you tell me the CPU % using this feature is still around 1x depth sensor, then we can just stop there. If STVL with this is no heavier than STVL with a depth  camera, then this is at the point of being moot. 

Your idea seems sound, how do you plan to calculate the right order? 

---

PS I will also want this PR on the `ros2` branch so new distribution users can have this as well, not just Jazzy. But don't worry about that until we're done reviewing the code.

remco-r:
TLDR; I discovered and committed some more optimizations. And I realized some downsides of using the 3D half plane representation.

### 1 SoA optimization

I had a suspicion about the previous optimization efforts in regards to the "flat hierarchy". It turns out I did something wrong in keeping track of all the different benchmarks last time. Now i got significant improvements by having the polygons in a structure of arrays (SoA) layout with contiguous vectors. Benchmarks consistently showed an approximate **halfing of the cpu time** (e.g. 201->90.8us) for `isInsideAnyObstruction`, and an overall **20% speed up** of the full `isInside` method. I **committed** this optimization. For the best cpu caching optimization it does not do early return anymore, but checks all polygons.

<details>

<summary> Benchmark results </summary>


```
Benchmark                                             Time             CPU   Iterations UserCounters...
-------------------------------------------------------------------------------------------------------
IsInsideBenchmark/IsInsideBatch/1/100000/0/0       2017 us         2014 us          356 items_per_second=49.6637M/s points=100k polygons=1 soa=0
IsInsideBenchmark/IsInsideBatch/1/100000/0/1       1610 us         1610 us          437 items_per_second=62.1172M/s points=100k polygons=1 soa=1
IsInsideBenchmark/IsInsideBatch/2/100000/0/0       2374 us         2374 us          294 items_per_second=42.1248M/s points=100k polygons=2 soa=0
IsInsideBenchmark/IsInsideBatch/2/100000/0/1       1861 us         1860 us          377 items_per_second=53.7505M/s points=100k polygons=2 soa=1
IsInsideBenchmark/IsInsideBatch/3/100000/0/0       2672 us         2672 us          262 items_per_second=37.4317M/s points=100k polygons=3 soa=0
IsInsideBenchmark/IsInsideBatch/3/100000/0/1       2115 us         2114 us          332 items_per_second=47.2947M/s points=100k polygons=3 soa=1
PolygonMicroBench/Test/4/1/0/0                      201 us          201 us         3505 edges=4 items_per_second=99.6696M/s near_miss=0 soa=0
PolygonMicroBench/Test/4/1/1/0                     90.8 us         90.8 us         7617 edges=4 items_per_second=220.218M/s near_miss=0 soa=1
PolygonMicroBench/Test/8/1/0/0                      241 us          241 us         2920 edges=8 items_per_second=83.002M/s near_miss=0 soa=0
PolygonMicroBench/Test/8/1/1/0                      105 us          105 us         6714 edges=8 items_per_second=191.263M/s near_miss=0 soa=1
PolygonMicroBench/Test/16/1/0/0                     258 us          258 us         2726 edges=16 items_per_second=77.4498M/s near_miss=0 soa=0
PolygonMicroBench/Test/16/1/1/0                     125 us          125 us         5633 edges=16 items_per_second=159.784M/s near_miss=0 soa=1
```
</details>

### 2 Polygon ordering

I also implemented a heuristic to test "largest area polygon first" to have the highest probability of early return between testing polygons. This is done by sorting in the pre computations on initialization. The savings are highly dependent on polygon layout. When having a huge one and a tiny one, sorting will have a big impact. The resulting improvements should be same or better. Benchmarked for several relevant polygon layouts, and it sped up `isInsideAnyObstruction` runtime 2 to 4 times. (e.g. 466->226ms). Also **committed**. 

### 3 Downsides of 3D half plane representation

I figured out that the new polygon edge representation as 3D half planes is not a one on one translation from the 2D azimuth elevation representation. To illustrate this i drew a couple of vertices in the gui and project the half plane representations of the edges back into 2D. 

Screenshot from 2026-07-24 15-48-11

_The green edges between the vertices "bend" more the further they are from the "equator"_

Now, to know what is inside and outside of the polygon when drawing them up, the user actually needs to see this half plane projection. So it becomes a bit more complicated for user friendliness. Also, convexity does not always translate from 2D to 3D case, so i had to update the check in `preCompute` to check for spherical convexity (**committed**). The convexity is necessary to guarantee the `isInside` normal cross product calculation is correct. I dont know yet how the final user will create the polygons, but the gui i used should also be updated to decompose into spherically convex polygons.

chart_vs_sphere_convexity

_An extreme example to show convexity does not translate between representations_

### 3 Ordering of edge normal checks

I had some good ideas for the ordering of edge normal checks in the 2D case. However, I abandoned this effort, because it becomes quite complex with the 3D half circles. Also the early return in `isInside`, which was the reason for efficiency gains, is currently optimized away in **1**.

### 4 Testing load with other sensors

I have no depth camera, but i can easily simulate one. Is it about using the other frustum, or because of the resolution of the sensor?

(I know you ask me because you want to compare loads of the full layer for different use cases, because if 3DLidar is already computationally below high res depth camera, optimizations make less sense, but i still want to mention the following: Technically the resolution of the sensor should not matter for this branch. The isInside check runs for each voxel, not for each point of the sensor, and the associated load is thus only dependent on voxel map size, voxel resolution and frustum size. But you probably knew this already)

What is the test you would like to see? What kind of depth sensor vs what kind of lidar?

### 5 Side quest

Changing the voxel coordinate transform to not use quaternion also ~halfs the computational load of this line, and speeds up `isInside` with ~20%. It's kind of out of the scope of this PR, but I could add it, or make another PR.
```
  - Eigen::Vector3d transformed_pt = _orientation_conjugate * (point_in_global_frame - _position);
  + Eigen::Vector3d transformed_pt = _rotation * point_in_global_frame + _translation;
```

SteveMacenski:
So where are we at in total with the perf in terms of before and after this PR with those adjustments? Sounds like we're probably in the ballpark! 

I'll skip reviewing the actual main file until we're settled, but I looked over everything else and I approve the glue/piping you added. 

> Changing the voxel coordinate transform to not use quaternion also ~halfs the computational load of this line, and speeds up isInside with ~20%. It's kind of out of the scope of this PR, but I could add it, or make another PR.

:eyes: is this a good overall system gain?

README.md
```diff
@@ -150,6 +150,23 @@ rgbd_obstacle_layer:
...
+    clearing: true
+    model_type: 1                #3D Lidar
+    horizontal_fov_angle: 6.28   #radians, full circle
+    vertical_fov_angle: 0.52     #radians
+    decay_acceleration: 5.0
+    obstruction_polygons: "[[0.8,-0.2, 1.2,-0.2, 1.2,0.1, 0.8,0.1], [3.5,-0.3, 4.0,-0.3, 4.0,0.0, 3.5,0.0]]"
+```
+
+Each inner bracket `[az1,el1, az2,el2, ...]` defines one convex polygon (minimum 3 vertices, azimuth in [0, 2π] radians, elevation in [-π/2, π/2] radians). If empty or unset, behavior is identical to the original STVL. Only applies to `model_type: 1`.
+
```
SteveMacenski, line 172: Adding your latest green Identifying obstruction voxels gif here might be nice to visualize!

remco-r: Added gif

remco-r:
Some highover numbers from newly run benchmarks and perf analysis:

**Benchmark IsInside()** with 100.000 points without polygons and with 2 polygons that together obstruct about 20% FOV:
* Original: no polygons (is equal to before PR) -> ~1200 us
* Polygons old: 2 forklift mast like polygons, 2D polygon handling -> ~4650 us
* Polygons optimized: 2 forklift mast like polygons, 3D vector optimization, flat vector SoA optimization -> ~ 1950 us

Conclusion: reasonable amount of polygons increases IsInside() runtime ~3.9x, with optimizations its ~1.6x.

> 👀 is this a good overall system gain?

**Benchmark no quaternion math:**
* No polygons: -> ~800 us
* Polygons old: 2 forklift mast like polygons, 2D polygon handling -> ~4100 us
* Polygons optimized: 2 forklift mast like polygons, 3D vector optimization, flat vector SoA optimization -> ~ 1550 us

Conclusion: removing per point quaternion math makes IsInside() run 11-33% faster. Minimal change; low hanging fruit.

**overall perf of the layer**
Trying to separate layer cpu% from total costmap exe cpu% is difficult. But by far the biggest two entrypoints are `spatio..::PointCloud2Callback (22%)`  and `spatio..::updateBounds (~17%)`. After removing the quaternion math  `spatio..::updateBounds` used ~14%. This gained only 3% of the total executable, but 7% of the layer and 17% of updateBounds. updateBounds' performance is still very related to how many voxels exist, but this gives a ballpark.

Calculating isInside the same way: after all optimizations it uses ~1% of layer cpu (or ~3% of updateBounds) in the forklift obstructed use case.

SteveMacenski:
> Conclusion: reasonable amount of polygons increases IsInside() runtime ~3.9x, with optimizations its ~1.6x.

OK, lets use that then! 

> Conclusion: removing per point quaternion math makes IsInside() run 11-33% faster. Minimal change; low hanging fruit.

What math are you using instead and does that suffer from singularity issues? I don't think I've actually seen the implementation or strategy. But this does sound relatively small. Lets do this if there's no downsides :-)

remco-r:
Cool!
* I cleaned up the obstruction_polygons.hpp since it became a bit convoluted after several updates.
* final commit implements rotation with rotation matrix to avoid quaternion calculations. Should be singularity free.
* updated https://github.com/remco-r/spatio_temporal_voxel_layer/tree/feature/obstruction_tests 
  * with tests for 3d half plane representation. 
  * and updated vibe coded gui that now handles 3d half plane visualizations and convexity decompositions.
    * gui should be installed with colcon and can be run with `python3 -m fov_boundary_tool.main`

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_grid.cpp
```diff
@@ -128,9 +128,13 @@ void SpatioTemporalVoxelGrid::ClearFrustums(
         it->_vertical_fov_in_rad,
         it->_horizontal_fov_in_rad, it->_min_z_in_m, it->_max_z_in_m);
     } else if (it->_model_type == THREE_DIMENSIONAL_LIDAR) {
-      frustum = new geometry::ThreeDimensionalLidarFrustum(
+      auto * lidar_frustum = new geometry::ThreeDimensionalLidarFrustum(
```
SteveMacenski: Why the temporary object?

spatio_temporal_voxel_layer/src/frustum_models/three_dimensional_lidar_frustum.cpp
```diff
@@ -78,8 +83,7 @@ bool ThreeDimensionalLidarFrustum::IsInside(const openvdb::Vec3d & pt)
 /*****************************************************************************/
 {
   Eigen::Vector3d point_in_global_frame(pt[0], pt[1], pt[2]);
-  Eigen::Vector3d transformed_pt =
-    _orientation_conjugate * (point_in_global_frame - _position);
+  Eigen::Vector3d transformed_pt = _rotation * point_in_global_frame + _translation;
```
SteveMacenski: I don't really mind, but does this actually change anything perf wise?

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
```diff
@@ -77,11 +83,15 @@ class ThreeDimensionalLidarFrustum : public Frustum
   double _min_d_squared, _max_d_squared;
   double _tan_vFOVhalf;
   double _tan_vFOVhalf_squared;
+  // externally set postion and orientation of sensor in global space
   Eigen::Vector3d _position;
   Eigen::Quaterniond _orientation;
-  Eigen::Quaterniond _orientation_conjugate;
+  // computed transform matrices to apply to voxels during clearing
+  Eigen::Matrix3d _rotation;
+  Eigen::Vector3d _translation;
```
SteveMacenski: We have `_position` and `_orientation` above, so maybe these need more specific names to differentiate?

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
```diff
@@ -0,0 +1,478 @@
...
+    }
+    if (is_finite) {
+      RCLCPP_WARN(logger, "Obstruction polygon %zu has NaN/inf vertices, skipping.", idx);
+      continue;
+    }
+
+    // Verify azimuth within bounds [0-2pi]
+    bool out_of_range = false;
+    for (const auto & v : vertices) {
+      if (v.azimuth < 0.0 || v.azimuth > 2.0 * M_PI) {
```
SteveMacenski, line 326: Any elevation check make sense?

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
```diff
@@ -0,0 +1,478 @@
...
+      const double len_sq =
+        normals[i].x * normals[i].x + normals[i].y * normals[i].y + normals[i].z * normals[i].z;
+
+      constexpr double kMinEdgeNormSquared = 1e-12;  // reject edges < ~0.001° apart
+      if (len_sq < kMinEdgeNormSquared) {
+        return std::nullopt;
+      }
+    }
+
+    // Determine orientation: centroid direction should be on inside of all half-planes
```
SteveMacenski: From AI:

centroid orientation is not robust for arbitrary spherical polygons

This:

Vec3D centroid{0.0, 0.0, 0.0};
for (const auto & d : dirs) {
  centroid.x += d.x;
  centroid.y += d.y;
  centroid.z += d.z;
}

and then:

const double test_dot =
  normals[0].x * centroid.x +
  normals[0].y * centroid.y +
  normals[0].z * centroid.z;

if (test_dot < 0.0) {
  ...
}

implicitly assumes that the vertex-sum vector points toward the interior of the spherical polygon.

That's generally reasonable for the small, localized spherical polygons you presumably want for sensor blind spots, but it is not universally true.

For polygons approaching or exceeding a hemisphere, the vertex sum can:

approach zero,
point toward the wrong side,
or become a poor representation of the polygon interior.

If your polygons are guaranteed to be small FOV obstruction regions, this is probably acceptable. But the API currently claims generic spherical convexity.

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
```diff
@@ -0,0 +1,478 @@
...
+
+/**
+ * @brief Parse a list of polygons from a footprint-style string.
+ *
+ * Format: "[[x1,y1, x2,y2, x3,y3], [x4,y4, x5,y5, x6,y6]]"
+ * Each inner [...] is one polygon with flat x/y vertex pairs.
+ *
+ * @return vector of parsed angular-domain polygons
+ */
+inline std::vector<AngularPolygon> parsePolygonsFromString(
```
SteveMacenski: We should fail parsing on any parsing error rather than taking partial results, that's a safety issue

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
```diff
@@ -0,0 +1,478 @@
...
+    if (test_dot < 0.0) {
+      for (auto & norm : normals) {
+        norm.x = -norm.x;
+        norm.y = -norm.y;
+        norm.z = -norm.z;
+      }
+    }
+
+    // Verify spherical convexity. For each edge all vertices must lay on same side
+    constexpr double kConvexEps = 1e-9;  // tolerance for vertices lying on an edge's line
```
SteveMacenski: From AI: 

Normals are not normalized, which makes every tolerance scale dependent. kConvexEps = 1e-9 is compared against dot(n_i, d_k) where |n_i| = sin(edge span). For a 1e-3 rad edge the effective angular tolerance is ~1e-6 rad; for a 1 rad edge it is ~1e-9 rad. Normalize the normals in double before the float cast. That also fixes the float precision loss on narrow edges and gives you a meaningful physical quantity: with a normalized normal and a normalized point, the dot is the sine of the angular distance to the edge plane.

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
```diff
@@ -0,0 +1,478 @@
...
+
+  for (size_t idx = 0; idx < input.size(); ++idx) {
+    const auto & vertices = input[idx];
+    if (vertices.size() < 3) {
+      RCLCPP_WARN(logger, "Obstruction polygon %zu has < 3 vertices, skipping.", idx);
+      continue;
+    }
+
+    // Check for NaN/inf
+    bool is_finite = false;
```
SteveMacenski: I think this logic is inverted; shouldn't it be true and switched to false; then checked if false?

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
```diff
@@ -0,0 +1,478 @@
...
+        logger, "Parsed %zu obstruction polygon(s) for %s", raw_polygons.size(),
+        param_prefix.c_str());
+    }
+
+    auto valid_polygons = validatePolygons(raw_polygons, logger);
+    if (valid_polygons.empty()) {
+      return nullptr;
+    }
+
+    auto filter = std::make_shared<ObstructionFilter>();
```
SteveMacenski: Rename: obstruction_filter. This would cause a compilation issue with build warnings

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
SteveMacenski: Winding detection is fragile. Orientation is decided by one dot product of normals[0] (unnormalized) against a non-normalized vertex-sum centroid. For a polygon roughly symmetric about that first edge's plane, test_dot approaches zero and the sign is decided by roundoff. The convexity check downstream makes this fail safe rather than silently inverted, but the user gets a misleading "non-convex or degenerate" rejection on a perfectly good polygon. Decide winding from the summed dot over all normals, or from the signed spherical area, and reject explicitly if the magnitude is below threshold.

SteveMacenski (commented):
A couple of pieces of commentary from AI - since I'm a little short on time and the obstruction polygons file is complex 😉 

Obstruction is purely angular with no range component. Voxels closer than the occluder inside the wedge are observable but will never be frustum cleared. This is fail safe but means stale obstacles in front of the mast persist for the full voxel_decay. A per-polygon min range would fix it properly; at minimum document it.

If the same source marks and the polygon covers self-returns, those voxels are now permanently exempt from frustum clearing. With voxel_decay: -1 that makes the robot body a permanent obstacle in the costmap. Worth an explicit README warning to pair this with height filtering or a self filter.

nickovaras:
This is very neat!

SteveMacenski:
Happy to have your review @nickovaras !

remco-r:
Those were good comments. I spend some time rewriting and catching more failure cases in the 3D half plane representation.
 I realized that edges spanning the azimuth at the 2pi seam are actually valid half circle edges. This means any edge longer than half a circle will flip the other way. I added documentation for this. 
Also added more documentation making the user aware of the half plane representation bowing the edges and using this image for that.

obstruction_polygon_projection

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_grid.cpp
remco-r: This was necessary to access the method setObstructionPolygons which was not accessible from the base class. However, I updated it now so that its passed in the constructor, now it looks exactly like the other frustum.

spatio_temporal_voxel_layer/src/frustum_models/three_dimensional_lidar_frustum.cpp
remco-r: This is actually the "last commit" perf optimization i talked about earlier. By using `_rotation` in stead of `_orientation_conjugate`, the use of `Eigen::Quaterniond::operator*(const Vector3d&)` is avoided which reduced the runtime of this line over many points significantly.

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/frustum_models/three_dimensional_lidar_frustum.hpp
remco-r: you are completely right, renamed them to something more meaningful

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
remco-r: indeed, elevation check makes equal sense since user may give something outside of [-pi/2,pi/2] which would be invalid in 2d az/el definition space. added check.

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
remco-r: updated such that exceptions are thrown. this will cause a failure during costmap onConfigure like "costmap: Failed to initialize costmap plugin..."

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
remco-r: updated with normalization

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
remco-r: thanks for catching that one. with newer exception throwing the variable is now not used anymore and throw happens directly in if clause that checks finiteness

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
remco-r: renamed

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
remco-r: * polygon edge case handled now, including check if vertices are on same hemisphere.
* winding detection is completely updated. now there are several checks for norm size, dot product sign and solid_angle size. Should be good now.

spatio_temporal_voxel_layer/src/frustum_models/three_dimensional_lidar_frustum.cpp
```diff
@@ -112,6 +122,17 @@ bool ThreeDimensionalLidarFrustum::IsInside(const openvdb::Vec3d & pt)
...
+  // Check if the point falls inside any obstruction polygon (blind spot)
+  if (_obstruction_filter) {
+    if (_obstruction_filter->isObstructed(
+        static_cast<float>(transformed_pt[0]),
+        static_cast<float>(transformed_pt[1]),
+        static_cast<float>(transformed_pt[2])))
+    {
+      return false;  // point is in a blind spot
+    }
+  }
```
SteveMacenski:
```suggestion
  if (_obstruction_filter && _obstruction_filter->isObstructed(
      static_cast<float>(transformed_pt[0]),
      static_cast<float>(transformed_pt[1]),
      static_cast<float>(transformed_pt[2])))
  {
    return false;  // point is in a blind spot
  }
```

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
```diff
@@ -0,0 +1,569 @@
...
+              "covers a near-zero solid angle (" + std::to_string(solid_angle) +
+              " sr) and would mask nothing");
+    }
+    if (solid_angle > kMaxSolidAngle) {
+      throw std::runtime_error(
+              "covers " + std::to_string(solid_angle) +
+              " sr, far too large for a blind spot; check whether an edge's azimuth step exceeds"
+              " pi and sweeps the wrong way, and slice it into several polygons");
+    }
+
```
SteveMacenski, line 271: I'm spouting AI on this - sorry I usually try to give things reviews from my own eyeballs not just throwing into Claude, but I'm a bit short on time this week :-) 

I am parsing this down to be more human readable and suggested actions though, which I hope makes up for it 😉 

---

A polygon with a >π azimuth edge silently produces the inverted mask, and validation accepts it. The kMaxSolidAngle = M_PI check is meant to catch edges that sweep the wrong way, but it does not.

Reproduction case: "[[0.0,-0.1, 3.5,-0.1, 3.5,0.1, 0.0,0.1]]":

* accepted, solid angle 2.01 sr, under the π limit
* isObstructed returns true for az 3.5 through 6.0 and false for az 0.5 through 3.0

So the user gets everything except the band they drew. kMinCentroidNorm = 1e-9 (L188) does not catch it. The centroid norm here is ~0.36, nowhere near the threshold, so that guard only fires on exactly antipodal vertex sets.

A potential (AI, to validate) Fix: check the azimuth step per edge inside the fromAngularVertices edge loop (L155-177), since that is the only place the user's intent is still recoverable:

```cpp
constexpr double kMaxEdgeAzimuthStep = M_PI_2;
const double daz = std::remainder(vertices[j].azimuth - vertices[i].azimuth, 2.0 * M_PI);
if (std::fabs(daz) > kMaxEdgeAzimuthStep) {
  throw std::runtime_error(
    "edge " + std::to_string(i) + " sweeps " + std::to_string(std::fabs(daz)) +
    " rad of azimuth; the geodesic takes the short way round, so split this into"
    " several polygons with edges under pi/2");
}
```

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
```diff
@@ -273,6 +274,15 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
           filter, voxel_min_points, enabled, clear_after_reading, model_type,
           node->get_clock(), node->get_logger())));
 
+    // Load obstruction polygons for 3D lidar sources
+    if (model_type == THREE_DIMENSIONAL_LIDAR) {
+      std::string param_prefix = name_ + "." + source;
+      auto filter = geometry::ObstructionFilter::fromParams(node, param_prefix, logger_);
```
SteveMacenski: Should `obstruction_polygons` be a dynamic parameter to change?

spatio_temporal_voxel_layer/src/frustum_models/three_dimensional_lidar_frustum.cpp
```diff
@@ -112,6 +122,17 @@ bool ThreeDimensionalLidarFrustum::IsInside(const openvdb::Vec3d & pt)
     }
   }
 
+  // Check if the point falls inside any obstruction polygon (blind spot)
```
SteveMacenski, line 125: A voxel one meter in front is observable but is now permanently exempt from frustum clearing and only decays temporally. A per-polygon min_range compared against radial_distance_squared before the isObstructed call is the proper fix and is nearly free since the value is already computed. 

Or, the 3D shapes defined need to have some non-sensor origin defined capabilities so that it can represent that it can see things closer than X distance and everything after an occlusion is not OK.

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
```diff
@@ -273,6 +274,15 @@ void SpatioTemporalVoxelLayer::onInitialize(void)
...
+      if (filter) {
+        _observation_buffers.back()->SetObstructionFilter(filter);
+      }
```
SteveMacenski:
```suggestion
      auto obs_filter = geometry::ObstructionFilter::fromParams(node, param_prefix, logger_);
      if (obs_filter) {
        _observation_buffers.back()->SetObstructionFilter(obs_filter);
      }
```

This shadows another parameter.

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
```diff
@@ -0,0 +1,569 @@
...
+  static std::shared_ptr<ObstructionFilter> fromParams(
+    NodeT node, const std::string & param_prefix, const rclcpp::Logger & logger)
+  {
+    std::string param_name = param_prefix + ".obstruction_polygons";
+    std::string polygons_str;
+    if (!node->has_parameter(param_name)) {
+      node->declare_parameter(param_name, std::string(""));
+    }
+    node->get_parameter(param_name, polygons_str);
+
```
SteveMacenski: You can use `Layer::declareParameter` as others do

spatio_temporal_voxel_layer/src/frustum_models/three_dimensional_lidar_frustum.cpp
SteveMacenski: ^ just flattening

SteveMacenski:
Generally LGTM!

spatio_temporal_voxel_layer/src/frustum_models/three_dimensional_lidar_frustum.cpp
SteveMacenski: I just might be mistaken here or something to explain to me / document that is unclear.

spatio_temporal_voxel_layer/src/frustum_models/three_dimensional_lidar_frustum.cpp
remco-r: applied

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
remco-r: I appreciate your effort of giving me the human experience :+1: . I dont mind a little AI here and there, especially for finding these edge case bugs, keeps me sharp. I will look at this later.

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
remco-r: short answer, yes, ill have a look into the implementation details.

The original design was for vehicle fixed obstructions. However a dynamic parameter would be convenient in the case of pickup up a load that temporarily becomes vehicle fixed and obstructs the view. Or if there is a large moving element like an end effector on the vehicle. When designing the obstruction filter, a lot of caching and processing was moved to the initialization of the class, so that the hotpath does as little operations as possible. I will have a look at the time the initialization takes, and other side effects of recreating the filter in the paramCallback.

spatio_temporal_voxel_layer/src/frustum_models/three_dimensional_lidar_frustum.cpp
remco-r: You're right. The design assumes you don't want to track obstacles within the vehicle footprint, and that these are filtered out before (which is the case for my setup). But indeed, anything observed between sensor and obstruction will now be added into the voxel map and not cleared until the time runs out or the vehicle moves. 
After giving it some thought, I still don't really see a use case for why someone would want to track and clear obstacles within the vehicle footprint. Especially since the obstruction filter only makes sense for moving vehicles, and the voxels are world fixed. However, since it's a small change i can add it. It will only be a bit funky having a second parameter vector with min_ranges_per_polygon.

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
remco-r: This is now called obstruction_filter (obs looked too much like observation) alongside a refactor of parameter loading (other conversation)

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
remco-r: I moved param loading out of the header, to where the other parameters are loaded, and followed that pattern. `fromParams` became a lot simpler, and the INFO logging is now done outside the header too.

spatio_temporal_voxel_layer/include/spatio_temporal_voxel_layer/obstruction_polygons.hpp
remco-r: Indeed it was posed that the "solid_angle > kMaxSolidAngle" check would be able to detect flipping edges, which is not possible from solid_angle alone. Neither is it possible to detect in any way, because it depends on knowing the intention of the user. The check was actually used as a plausibility check and should have only warned, not thrown an exception. 

My biggest fear was that it would not be intuitive for a user to indicate the obstruction polygons by azimuth/elevation when they are actually converted to half plane circles in the code, with all the associated flipping/bowing. 

I have decided to remove warnings like this, and to have a polygon visualizer in stead `publish_obstruction_markers`. This way, when configuring, the user can see the result in rviz. No need to warn for potential undesired effects, when the user can see the result in rviz.
It created some code overhead, but the result is quite satisfying:

stvl_obstruction_marker

spatio_temporal_voxel_layer/src/spatio_temporal_voxel_layer.cpp
remco-r: Implemented as a dynamic parameter. I also added min_ranges parameter, which is also dynamic. However, since they were tied, and rqt reconfigure only allows updating one parameter at a time, i had to allow temporary mismatch in size of polygons and min_ranges. Currently the user will be warned by mismatch. In any case a mismatch will cause min_range to be 0.

spatio_temporal_voxel_layer/src/frustum_models/three_dimensional_lidar_frustum.cpp
remco-r: implemented parameter `obstruction_min_ranges`

PR #360, opened 2026-08-19
Add configurable 3D lidar frustum orientation

ardavan34:
### Description:

During an internship, I used STVL with a LiDAR frame whose Y-axis pointed toward the front of the robot. Since the frustum assumes +X is forward, this inspired adding configurable roll, pitch, and yaw offsets.

This PR:
  - Preserves the existing +X-forward behavior by default
  - Supports arbitrary frustum orientation relative to the LiDAR frame
  - Adds documentation and unit tests for X, Y, intermediate, and tilted directions
  - Local geometry tests (passed)
  
PS: @SteveMacenski I've been inspired by your work and journey for the past couple of years. Just wanted to say thank you for everything you've done for the robotics community. This is my first open source contribution and I hope I get to learn a lot from this experience :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,10 +148,25 @@ rgbd_obstacle_layer:
     vertical_fov_angle: 0.7      #default 0.7, radians
     vertical_fov_offset: 0.0     # default 0, radians. 3D Lidar only. Offset from planar axis. f.e. MID360 has a vFOV of -7deg to 52deg => offset +22.5deg (0.3927 rad)
     horizontal_fov_angle: 1.04   #default 1.04, radians
+    frustum_roll: 0.0            #default 0, radians. 3D Lidar only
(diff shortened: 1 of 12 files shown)


SteveMacenski:
Thanks! 

REP 103 (https://www.ros.org/reps/rep-0103.html) highlights the proper coordinate frame system for all ROS compliant systems. It seems like your sensor is not properly setup that would cause this issue if Y was forward in the coordinate frame, and not simply installed offset by 90 degrees. I would suggest raising it with the ROS driver for that sensor manufacturer (or fixing it internally if its an internally developed driver) :-) 

Unfortunately as this breaks some very foundational convention, I don't think we can merge this one in - the fix for this belongs at the ROS driver level for the sensor

